PR #41, opened 2020-01-19
move packet_size to events, move packet_type to PacketHeader

### LINKED ISSUE #40: packet_size doesn't belong in the PacketHeader, but PacketType does
In my interpretation, the `PacketHeader` is a QUIC packet header. Therefore, it should contain the QUIC packet type.
However, the `packet_size` (as opposed to the `payload_length`, which is the length of the QUIC payload e.g. in a coalesced packet), is a property of the UDP packet, and therefore should a property of the `packet_sent` / `packet_received` event.

marten-seemann:
Fixes #40.

### DIFF
--- a/draft-marx-qlog-event-definitions-quic-h3.md
+++ b/draft-marx-qlog-event-definitions-quic-h3.md
@@ -495,7 +495,7 @@ Data:
 
 ~~~
 {
-    packet_type:PacketType,
+    packet_size?:number,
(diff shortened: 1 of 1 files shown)


rmarx:
This was a slightly bigger change than initially anticipated and made in https://github.com/quiclog/internet-drafts/commit/0cb1f02e3c8a60962bc39b387c5db9ba12210556.

PR #61, opened 2020-03-08
rename header_decrypt_error to header_parse_error

### LINKED ISSUE #60: packet_dropped header_decrypt_error should be header_parse_error
Header decryption can never fail: it's just an XOR with a byte mask (header protection is not authenticated). You'd drop a packet though if header parsing fails, so maybe we can rename the trigger?

marten-seemann:
Fixes #60.


rmarx:
Made change manually in draft02 branch in https://github.com/quiclog/internet-drafts/commit/a578e935e4cae5d3415bb46358a7b668bf4a4356

PR #68, opened 2020-03-25
Add stateless reset support

### LINKED ISSUE #64: Properly specify stateless reset, retry and migration
These concepts were in flux when -01 was made and are now much more settled.

Any thoughts on how to best approach this are more than welcome.

rmarx:
Partially fixes #64.

Main change is adding `stateless_reset` as a PacketType.

The `stateless_reset_token` is added as a field to packet_sent and packet_received.
Any of the "unpredictable bits" can be logged as `raw_encrypted`

Added a trigger to `connection_state_update` to indicate if a connection goes into draining/closing because a valid stateless_reset was received.

Backwards compatibility break: renamed `reset_token` to `stateless_reset_token` on the NewConnectionID frame definition (for consistency). 

Been thinking about a way to signal an "invalid" stateless reset, but figured there is no way to distinguish it from an otherwise garbage datagram, so decided it was impossible. Not 100% sure about the "looping case" where you continually lower the packet size. Maybe that's solved by adding a generic "packet_size" trigger value to `packet_dropped` though?

### DIFF
--- a/draft-marx-qlog-event-definitions-quic-h3.md
+++ b/draft-marx-qlog-event-definitions-quic-h3.md
@@ -357,6 +357,7 @@ Triggers:
 * "error" // when closing because of an unexpected event
 * "clean" // when closing normally
 * "application" // e.g., HTTP/3's GOAWAY frame
+* "stateless_reset" // because a stateless reset was received
 
 ### MIGRATION-related events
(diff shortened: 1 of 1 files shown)


kazu-yamamoto:
LGTM!

PR #73, opened 2020-03-28
Merge loss_timer events

### LINKED ISSUE #69: PTO is per PN-space
Currently, we log `pto_count` in `metrics_updated`. However, we can have multiple PTO's running across multiple PN spaces and there is no proper way to correlate them with a PN space in metrics_updated.

This also impacts the `loss_timer*` event types. 

Possible resolutions:
- Add pn_space attribute to `metrics_updated`
- Split PTO count into a separate event (potentially merge with loss_timer?)

TODO: reason more about how these metrics are tied to PN spaces in general (are there others than PTO that are duplicated)

cc @marten-seemann

rmarx:
Fixes #69.

Not 100% sure this is enough for all the latest draft-27(+) changes though, added a TODO in the draft for that purpose. 

Also added a packet_number_space field to metrics_updated to be flexible for implementations that don't want to implement all types of events. It's not a great fit at that location, but it seemed the most synergistic without having a completely new event type.

PTAL @marten-seemann

### DIFF
--- a/draft-marx-qlog-event-definitions-quic-h3.md
+++ b/draft-marx-qlog-event-definitions-quic-h3.md
@@ -817,8 +817,6 @@ Data:
 
     ssthresh?:number, // in bytes
 
-    packet_number_space?:PacketNumberSpace, // mainly necessary if an implementation does not log other events (e.g., key_retired) that indicate the pn-space has changed
-
     // qlog defined
     packets_in_flight?:number, // sum of all packet number spaces
     in_recovery?:boolean, // high-level signal. For more granularity, see congestion_state_updated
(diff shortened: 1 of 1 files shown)


draft-marx-qlog-event-definitions-quic-h3.md
```diff
@@ -866,41 +868,40 @@ Data:
...
-Data:
+Triggers:
 
-~~~
-{
-    timer_type?:"ack"|"pto", // called "mode" in draft-23 A.9.
-    packet_number_space?: PacketNumberSpace
-}
-~~~
+* no_packets_outstanding // for cancelling the timer
```
marten-seemann: Why do we need to export a trigger, if this is the only possible reason for canceling the timer anyway?

PR #77, opened 2020-04-10
add more packet_dropped triggers

marten-seemann:


rmarx:
Per discussion on slack, `unexpected_version_negotiation` and `unexpected_retry` can be merged into `unexpected_packet`, given that we have a `packet_type` field for the disambiguation.

marten-seemann:
@rmarx Done. Updated the PR.

PR #83, opened 2020-04-15
add a timeout trigger for the closed connection_state_updated event

marten-seemann:


marten-seemann:
I'm unsure about this PR. In my implementation, I have an idle timeout as well as a handshake timeout (the handshake timeout way shorter than the idle timeout).
Would it make sense to add a value for both? Or a separate field for trigger details?

rmarx:
In draft-02, I added a trigger value for both generic timeout and handshake_timeout based on quic-go's implementation.

PR #103, opened 2020-07-08
remove the UnknownFrame QUIC frame type

marten-seemann:
Other than H3 frames, QUIC frames are not TLV encoded. It's therefore not possible to parse a frame type that your implementation doesn't understand. In fact, it's a protocol violation to even send such a frame.

### DIFF
--- a/draft-marx-qlog-event-definitions-quic-h3.md
+++ b/draft-marx-qlog-event-definitions-quic-h3.md
@@ -1769,17 +1769,6 @@ class HandshakeDoneFrame{
 }
 ~~~
 
-### UnknownFrame
-
-~~~
-class UnknownFrame{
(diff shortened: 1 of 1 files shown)


rmarx:
I'm not sure I follow you on this one.

Here, the use case is for the receiver. They see an incoming frame with an unknown type, which, according to the spec, leads to a connection close:

> An endpoint MUST treat the receipt of a frame of unknown type as a
   connection error of type FRAME_ENCODING_ERROR.

This definition is to allow the endpoint to log more detailed info (or at least the raw value for manual dissection during debugging) about the unexpected frame before closing the connection.

marten-seemann:
That makes sense. Must have missed that, sorry.

PR #123, opened 2020-11-06
token-related fixes

### LINKED ISSUE #122: the new Token is confusing
The `Token` introduced in https://github.com/quiclog/internet-drafts/commit/60690f4033624006b5bbac8797e69b8a78331ce3 doesn't make sense to me. I don't see the motivation for bundling stateless resets and Retry tokens into the same struct (other than incidentally, the spec calls them tokens).

Stateless reset tokens are by definition random values (or rather, a deterministic pseudorandom value derived from the connection ID). They are always 16 bytes and never contain any other information.

On the other hand, the token sent on the Initial can be a Retry token or a NEW_TOKEN token (I guess that's what a resumption token is). This token typically does contain additional information encoded into the value (in fact, as I've described in https://github.com/quiclog/internet-drafts/issues/117#issuecomment-692603028, the spec mandates that you encode a bunch of values into that token). I would have wished that at least those mandatory fields are clearly defined in the `Token`.

marten-seemann:
Fixes #122.

This PR contains 2 fixes:

- stateless reset tokens are not tokens
- `PacketHeader` already contains a `Token`, so we don't need another field on the `packet_sent` and `packet_received` event

### DIFF
--- a/draft-marx-qlog-event-definitions-quic-h3.md
+++ b/draft-marx-qlog-event-definitions-quic-h3.md
@@ -670,7 +670,7 @@ Data:
     original_destination_connection_id?:bytes,
     initial_source_connection_id?:bytes,
     retry_source_connection_id?:bytes,
-    stateless_reset_token?:Token,
+    stateless_reset_token?:bytes,
     disable_active_migration?:boolean,
 
     max_idle_timeout?:uint64,
@@ -697,7 +697,7 @@ interface PreferredAddress {
     port_v6:uint16,
 
     connection_id:bytes,
-    stateless_reset_token:Token
+    stateless_reset_token:bytes
 }
 ~~~
 
@@ -2086,7 +2086,7 @@ class NewConnectionIDFrame{
   connection_id_length?:uint8;
   connection_id:bytes;
 
-  stateless_reset_token?:Token;
+  stateless_reset_token?:bytes;
 }
 ~~~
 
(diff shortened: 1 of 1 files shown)


draft-marx-qlog-event-definitions-quic-h3.md
```diff
@@ -1817,7 +1813,7 @@ class PacketHeader {
 
     flags?: uint8; // the bit flags of the packet headers (spin bit, key update bit, etc. up to and including the packet number length bits if present) interpreted as a single 8-bit integer
 
-    token?:Token; // only if packet_type == initial
+    token?:Token; // only if packet_type == initial || retry
```
rmarx, line 1816: So this is a bit of a difficult one imo. 
I'd put retry_token outside, since it's encapsulated in its own packet type, and so it can be considered to be part of the "payload" rather than the header. 
This is different from the initial token, which is clearly part of the header.

If we do put the retry token in the PacketHeader struct, imo it would make sense to do the same for the stateless_reset_token as well, since from the same logic, you could say that that is part of the Stateless Reset Packet header as well, rater than its payload. 

I think it's probably cleaner to have all three in the PacketHeader then, with token?:Token covering both initial and retry as you proposed here.

rmarx (changes requested):

marten-seemann:
Conclusion: remove retry_token from packet_sent and packet_received, allow using the token field on the header for Retry packets.

PR #125, opened 2020-12-03
Field name consistency with QUIC draft 32

toru:
Currently there's a mix of `limit` and `maximum` in the QLog spec, such that:

```
class DataBlockedFrame{
  frame_type:string = "data_blocked";

  limit:uint64;
}

...

 class MaxDataFrame{
   frame_type:string = "max_data";

   maximum:string;
 }
```

I was kindly informed in the QLog chatroom that this was inherited from `h3-27`. Fast forward to `h3-32` these fields have been unified to `maximum`. Therefore let's update the spec for consistency (with QUIC and within QLog itself).

### DIFF
--- a/draft-marx-qlog-event-definitions-quic-h3.md
+++ b/draft-marx-qlog-event-definitions-quic-h3.md
@@ -2047,7 +2047,7 @@ class MaxStreamsFrame{
 class DataBlockedFrame{
   frame_type:string = "data_blocked";
 
-  limit:uint64;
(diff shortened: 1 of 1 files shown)


rmarx:
Yeah, this is clearly an oversight.
Especially for this type of event/frame, the goal is to be fully consistent with the terminology in the QUIC texts. I just missed they re-named it there. 

Thanks for the PR!

Issue #75, opened 2020-03-30
Add way to log contents of version negotiation packet


rmarx:
Currently, there is no way to do this.

Possible approaches:
1. add `supported_versions:Array<string>` to `parameters_set` (similar to ALPN in #28)
2. add `supported_versions:Array<string>` to the `packet_*` events (similar to how we do stateless_reset

I personally have a preference for nr. 2, since it's consistent with how other packets are handled. ALPN is after all indeed a parameter, not a full packet. 

cc @mpiraux

marten-seemann:
I'd argue that we need both. The supported versions is a parameter of the server that's interesting to know independent of if version negotiation was ever performed or not.

For the client, it's important to have a way to log the contents of a version negotiation packet, so adding it to the `packet_received` event makes sense.

rmarx:
Related to #62

rmarx:
Option 2 was introduced in commit ac00848b50c1da3feaf50f43938e962a091396b0.

However, we currently still lack a way to indicate which versions each endpoint supports by themselves (which is not always communicated over the wire).

rmarx:
This was resolved for now by adding the new `version_information` event. Per #85, I'm not entirely sure that's the best approach here though. Closing this for now and keeping that issue open for general discussion. We can revisit this later depending on the outcome.

PR #138, opened 2021-03-15
introduce a VersionNegotiationError

marten-seemann:
When a client receives a Version Negotiation packet indicating that there's no commonly supported version, it aborts the connection attempt. Currently, qlog doesn't allow to log this.

### DIFF
--- a/draft-marx-qlog-event-definitions-quic-h3.md
+++ b/draft-marx-qlog-event-definitions-quic-h3.md
@@ -376,6 +376,7 @@ multiple error codes can be set on the same event to reflect this.
     connection_code?:TransportError | CryptoError | uint32,
     application_code?:ApplicationError | uint32,
     internal_code?:uint32,
+    version_negotiation: Array<bytes>,
 
     reason?:string
 }


marten-seemann:
Thinking about it, I'm not so sure about it any more. For stateless reset, we already use the `trigger` field. Maybe we should do the same for VN? Then we could close this PR.

rmarx:
I think my intention here was that:
- client logs `connection_closed` with trigger of value `version_mismatch` (which is already in -02, though potentially not perfectly named). Did you maybe miss that trigger? 
- to know which versions were offered and supported by both sides, you'd use the `version_information` event and `supported_versions` fields of `packet_sent/received` (see discussion in https://github.com/quiclog/internet-drafts/issues/75)

I agree this is slightly more tedious than having all of that together in a single event, but IIUC your last reply here, that isn't very critical.

rmarx:
@marten-seemann I'm not sure how to proceed on this one. From my perspective, my approach in the previous comment should be sufficient and this PR can be closed?

However, I did notice a few more aspects around this that I listed in #241. I suggest closing this PR and continuing discussion on that issue.

huitema (commented):
You asked me to look at this from the point of view of "compatible version negotiation". I think the compatible VN logging is already OK. Compatible VN negotiation uses transport parameters. These parameters are already logged, see for example:
```
[68563, "transport", "parameters_set", {
    "owner": "local",
    "initial_max_stream_data_bidi_local" : 2097152,
    "initial_max_data" : 1048576,
    "initial_max_streams_bidi" : 513,
    "idle_timeout" : 30000,
    "max_packet_size" : 1440,
    "initial_max_streams_uni" : 513,
    "initial_max_stream_data_bidi_remote" : 65635,
    "initial_max_stream_data_uni" : 65535,
    "active_connection_id_limit" : 8,
    "max_ack_delay" : 10,
    "handshake_connection_id": "091caa2b8bf8bc65",
    "1cce": "8000fcc7",
    "min_ack_delay" : 1000,
    "enable_time_stamp" : "03",
    "grease_quic_bit" : "",
    "version_negotiation": { "chosen": "00000001", "others": ["00000002", "00000001"]}}],
```

marten-seemann:
@rmarx Sorry for dropping the ball on this. After re-reading this:

> to know which versions were offered and supported by both sides, you'd use the version_information event and supported_versions fields of packet_sent/received (see discussion in https://github.com/quicwg/qlog/issues/75)

Or the `TransportPacketSent` with `header.packet_type == "version_negotiation"`.

> client logs connection_closed with trigger of value version_mismatch (which is already in -02, though potentially not perfectly named). Did you maybe miss that trigger?

Looks like I did.

I think this resolves the problem I was trying to solve, so we won't need this PR.

Issue #137, opened 2021-03-02
Prepare for adoption


rmarx:
List of thinks to keep in mind when preparing for QUIC wg adoption:
- Split QUIC and HTTP/3 event definitions into separate drafts (ensure proper cross-reference)
- Update editor names and affiliations
- Specify which value of `protocol_type` to use for each document (and change to array-based already?)
- Move shared stuff to the main schema document
  - RawInfo (though not 100% sure that's generic enough for all protocols?)
  - The whole "generic" category (error, warning, verbose, etc.)
  - Potentially the "Importance" indicators (Core, Base, Extra)
- Add more information about "previous versions" (e.g., two main versions in production, -01 and -02, those pre-date these adopted documents)
- Define a new version codepoint for the adopted drafts (now it's still draft-03-wip, but we also can't just do draft-00 or we'll overlap soon. Maybe qlog-0x?)
- rename master branch to main #140 
- Send e-mail to mailing list detailing the changes. Chairs will send a separate email with call for adoption afterwards.

Concrete steps for splitting / preparing the drafts according to @lpardue:

```
cp draft-marx-qlog-event-definitions-quic-h3.md draft-marx-qlog-quic-events.md
cp draft-marx-qlog-event-definitions-quic-h3.md draft-marx-qlog-h3-events.md

# add comment into draft-marx-qlog-event-definitions-quic-h3.md saying "beginning the split"

git add draft-marx-qlog-event-definitions-quic-h3.md draft-marx-qlog-quic-events.md draft-marx-qlog-h3-events.md

git commit -m "starting document split"

# then delete all the non-relevant text in draft-marx-qlog-quic-events.md and draft-marx-qlog-h3-events.md, and move necessary text to main-schema, then

git -rm draft-marx-qlog-event-definitions-quic-h3.md
git add draft-marx-qlog-quic-events.md draft-marx-qlog-h3-events.md

git commit -m "complete document split"

# then update editors and other metadata in the 3 documents

git commit -m "prepare for adoption"

# then submit 03, 00, 00 as individual I-Ds that we can issue a call for adoption on
```

LPardue:
I hadn't appreciated you might need to also move stuff into the main schema. If that is too onerous a task at this stage then you could just mark it in the docs and leave it there for the being.

The important thing is the get the overall structure in place and agreed. The content of each document can then be evolved through the working group.

PR #155, opened 2021-05-01
Prepare for adoption

rmarx:
This PR tracks the changes necessary to prepare for adoption by the QUIC wg as defined in #137.

It currently contains all the text changes, still requiring a few other things after they are approved (remove old document files, switch master branch to main, submit updated docs as individual drafts for the adoption call.


rmarx:
For those of you who don't use a local checkout of this to view the resulting HTML after the changes, I've attached them separately:

[pre-adoption-html-v1.zip](https://github.com/quiclog/internet-drafts/files/6409599/pre-adoption-html-v1.zip)

draft-marx-qlog-h3-events.md
```diff
@@ -0,0 +1,931 @@
...
+
+# Introduction
+
+This document describes the values of the qlog name ("category" + "event") and
+"data" fields and their semantics for the HTTP/3 and QPACK protocols. This
+document is based on draft-34 of the HTTP/3 I-D [QUIC-HTTP] and draft-21 of the
+QPACK I-D [QUIC-QPACK]. QUIC events are defined in a separate document
+[QLOG-QUIC].
+
+Feedback and discussion welcome at [TODO](TODO). Readers are advised to refer to
```
LPardue, line 129: You could still point people to this repo in order to collect anyion call feedback. Then, any and all issues can be brought across to the QUIC WG org when the repo is transferred. And we can update the URL to that new one when the adopted `-00` gets published.

draft-marx-qlog-event-definitions-quic-h3.md
```diff
@@ -73,6 +73,8 @@ informative:
 
 --- abstract
 
+ === beginning the split into two separate documents for quicwg adoption ===
```
LPardue, line 76: Is the intent to delete this file?

rmarx:
Two other points raised by Lucas:
- Add more information about "previous versions" (e.g., two main versions in production, -01 and -02, those pre-date these adopted documents)
- Define a new version codepoint for the adopted drafts (now it's still draft-03-wip, but we also can't just do draft-00 or we'll overlap soon. Maybe qlog-0x?)

draft-marx-qlog-event-definitions-quic-h3.md
rmarx: Indeed, this one is removed and replaced by the two new ones.

LPardue: so uhh, why isn't it `git rm`'d ? :D

LPardue:
Do we  also need independent versioning of the different schema? How does qlog evolve over the years? For example, say the main schema got published as 1.0, and then work commenced on 2.0. Can there be a QUIC qlog event definitions 1.1 (which is updated say to support QUIC v2) and a 2.1 (which is updated to support QUIC v2 and a new main schema format)? 

There's a lot of possible bikeshedding here. For now, maybe you just want to keep your linear versioning scheme (qlog 01, 02, and next 03) and make it clear that it is decoupled from any document draft id version.

marten-seemann (approved):
LGTM, but as @LPardue pointed out, don't we need to delete the old documents?

draft-marx-qlog-main-schema.md
```diff
@@ -866,6 +890,285 @@ trace has a different value for a given field, this field MUST NOT be added to
...
+
+This document only defines the main schema for the qlog format. This is intended
+to be used together with specific, per-protocol event definitions that specify the
+name (category + type) and data needed for each individual event. This is with the
+intent to allow the qlog main schema to be easily re-used for several protocols.
+Examples include the QUIC event definitions [QLOG-QUIC] and HTTP/3 and QPACK
+event definitions [QLOG-H3].
+
+This section defines some basic annotations and concepts the creators of event
+definition documents should follow to ensure a measure of consistency, making it
```
lnicco, line 903: is this a SHOULD ? not sure

rmarx:
Thanks all for the reviews! 

@lnicco I fully agree that the current text isn't really fit for a proper I-D and especially eventual RFC. I'm hoping we can make things much stricter and clearer following wg consensus on some main points. I'm also hoping the new editors will help re-write some things when that happens ;) 


I have just fixed the remarks (hopefully), merged the PR and tried to submit the new drafts to the datatracker. 
I ran into some problems due to not being familiar with the process there though, doing something different for all 3 documents...

- main-schema-03 should be published as normal (was somehow auto-posted to the QUIC list...)
- quic-events-00 somehow ended up as having to go through a manual review process (editors got an email about that)
- h3-events-00 has to be approved by the QUIC wg chairs first (no clue why/how that triggered, but @LPardue got an email for that, so...)

So you'll probably see some weird emails and the 2 event-definition documents are somewhat delayed behind main-schema, but hopefully things will be sorted soon!

rmarx:
After guidance from the QUIC wg chairs, I have now cancelled submission of:
- draft-marx-qlog-quic-events-00
- draft-marx-qlog-h3-events-00

And instead re-submitted as:
- draft-marx-quic-qlog-quic-events-00
- draft-marx-quic-qlog-h3-events-00

to make it clearer these are adoption candidates. These are both still pending approval from the QUIC wg chairs, so not quite published yet.

rmarx:
The soap continued... apparently this all triggered a bug in datatracker when trying to replace documents that are in pre-adoption with a wg (it's not quite a compiler bug, but I'll take it for now ;)). 

The chosen solution is to submit the new drafts again, without indicating they replace the old joint h3+quic events draft. The idea is that the backlinking will be corrected post-adoption.

For clarity, these are the most recent drafts that the adoption call will be issued for:
- https://datatracker.ietf.org/doc/draft-marx-quic-qlog-h3-events/
- https://datatracker.ietf.org/doc/draft-marx-quic-qlog-quic-events/
- https://datatracker.ietf.org/doc/draft-marx-qlog-main-schema/

PR #177, opened 2021-09-20
Fix 0RTT and 1RTT definitions in PacketType

### LINKED ISSUE #128: Remove implementation detail from PacketType enum
At the moment, the PacketType enum has `onertt = "1RTT"` (and similar for 0RTT) because in most programming languages, enum keys cannot start with numerical characters. This has however led to some confusion in some cases, so it's probably best to leave out this implementation detail in the spec.

See also https://github.com/p-quic/pquic/issues/18#issuecomment-756005163

lnicco:
Fixes #128

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1300,8 +1300,8 @@ class IPAddress : string | bytes;
 enum PacketType {
     initial,
     handshake,
-    zerortt = "0RTT",
-    onertt = "1RTT",
+    0RTT,
+    1RTT,
     retry,
     version_negotiation,
     stateless_reset,


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1300,8 +1300,8 @@ class IPAddress : string | bytes;
 enum PacketType {
     initial,
     handshake,
-    zerortt = "0RTT",
-    onertt = "1RTT",
+    0RTT,
```
lnicco, line 1303: nit: one could argue that RTT should be lowercase for consistency with the other types.

rmarx: Somewhat disagree. RTT is an acronym and the others are not. The RFCs also use the uppercased version of this. Finally, for backwards-compat with existing qlog implementations, 0RTT and 1RTT is easiest.

lnicco: @rmarx sounds good. 
Are there any other concerns, or can this be merged?

rmarx: This is fine I think. 

For other, larger changes, we'd ideally also add an entry in the changelist on the bottom. But this is such a small thing that it's not needed.

Issue #188, opened 2022-02-14
Force line length to 72


rmarx:
According to the official IETF draft guidelines, line length should not be longer than 72 ASCII characters (https://www.ietf.org/how/ids/guidelines/#format, and https://www.ietf.org/how/ids/guidelines/).

For some reason, I had previously understood this to be 82 characters, and so the current texts are 10 characters too wide...

Since this will be somewhat of a tedious process (which can be done largely automatically but with manual checks) which will impact allmost all parts of the documents, I propose we do this in a single burst somewhere down the line. 

cc @LPardue

rmarx:
This should be fixed as of main-schema-02 and h3/quic-events-01. 

For future reference: hard limit of 72 is only needed inside "figures" (including code blocks!). Running text paragraphs/lists/etc. can be done by the tooling itself.

PR #189, opened 2022-02-14
Fix main schema build errors and wrapping

LPardue:
- main: temporarily remove {: .language-json} tags until diagrams are changed
- main: wrap lines to appease xml2rfc

Helps address part of #188.

In this spec, you have a fundamental problem that the json in examples is malformed and kramdown complains. So let's just disable that for now. 

The rest of the changes are tedious but also not too onerous. The main issue you have is that there's lots of comments here and there that break length limits. So I've just picked a line length that will appease xml2rfc. 

In some cases, you have comments inside JSON, which won't work. In one case I deleted some because the qlog properties might get deleted anyway.

The changes you have in mind might have obviated the errors anyway. But if not, these changes will set a strong foundation to know how to shut xml2rfc up.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -244,7 +244,6 @@ JSON serialization example:
     "traces": [...]
 }
 ~~~~~~~~
-{: .language-json}
 {: #top-element title="Top-level element"}
 
 ## summary
@@ -282,7 +281,6 @@ JSON serialization:
     "error_count": 2
 }
 ~~~
-{: .language-json}
 {: #summary-example title="Summary example definition"}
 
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1477,11 +1481,10 @@ JSON-SEQ serialization example:
       }
     }
 }
-<RS>{"time": 2, "name": "transport:packet_received", "data": { ... } }
-<RS>{"time": 7, "name": "http:frame_parsed", "data": { ... } }
+<RS>{"time": 2, "name": "transport:parameters_set", "data": { ... } }
+<RS>{"time": 7, "name": "transport:packet_sent", "data": { ... } }
```
nibanks, line 1485: Why these changes?

LPardue: in this specific case, the newlines matter. While the format does support them, it might confuse readers. So I changed the events to ones that had a slightly shorter name but still seemed to be semantically valid.

nibanks: Ok.

nibanks (approved):

rmarx (approved):

PR #193, opened 2022-02-19
Move to CDDL - main schema

rmarx:
This PR bundles the TypeScript-to-CDDL changes for draft 0.4 for the main schema document (see other PRs for the QUIC and H3/QPACK documents). 

List of intended changes:

- [x] Replace TypeScript with CDDL descriptions
- [x] Split up CDDL definitions and JSON examples
- [ ] Properly explain the $ProtocolEventBody extension point + include trigger in the example 
- [ ] Properly discuss how event docs should define triggers in the CDDL 
- [ ] Replace TypeScript explanation in `Notational Conventions` and `qlog to JSON mapping` with CDDL and I-JSON references, basic explanations and considerations
- [ ] Explain the need for custom uint64 and hexstring CDDL types

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -176,6 +176,25 @@ doubles in the millisecond resolution.
 Other qlog documents can define their own data types (e.g., separately for each
 Packet type that a protocol supports).
 
+
+qlog also defines two custom CDDL types:
+
+Definition:
+
+~~~~~~~~
+; CDDL doesn't define a default uint64 type
+; NOTE: this is below QlogFile here because otherwise the tooling
+; thinks uint64 is the root type instead of QlogFile...
+uint64 = uint .size 8
+; CDDL doesn't define a proper hex-string that's compatible with JSON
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -217,22 +236,26 @@ In order to make it easier to parse and identify qlog files and their
...
+QlogFile = {
+    qlog_version: text
+    ? qlog_format: text .default "JSON"
+    ? title: text
+    ? description: text
+    ? summary: Summary
+    ? traces: [+ Trace / TraceError]
 }
+~~~~~~~~
+{: .language-cddl}
```
lnicco, line 253: in #194 I have used `~~~~ cddl` to open the block instead of this line. 
It seems to generate the same HTML, it's a bit less verbose and it also is easier to parse with some simple awk so I have found it to be quite handy of a syntax compared to what you are using here.

It would be nice to keep things consistent.

rmarx: I agree your approach is probably the better option (and potentially makes it easier to extract CDDL from the markdown docs), so I'll change my side. Thanks!

PR #203, opened 2022-05-16
Fix #202: Add PRIORITY_UPDATE support

### LINKED ISSUE #202: Add support for HTTP/3 PRIORITY_UPDATE
This will be RFC soon, I'd like for the HTTP/3 schema to include this event formally. Spec: https://httpwg.org/http-extensions/draft-ietf-httpbis-priority.html#section-7.2

LPardue:
Adds support for HTTP/3 PRIORITY_UPDATE frame.

Closes #202


lnicco (approved):
Good catch! 
Looks good to me.

PR #207, opened 2022-06-29
fix regex for TLS alert codes

### LINKED ISSUE #200: TLS alert error code is larger
Currently, we define the TLS_ALERT error codes as 0x100 to 0x199, while apparently in RFC9000 it's [0x0100, 0x01ff] (see https://datatracker.ietf.org/doc/html/rfc9000#section-22.5).

Should be tackled as part of #174.

marten-seemann:
Fixes #200.

TLS alerts are from 0x100 - 0x1ff, not 0x100 - 0x199.


lnicco (approved):
this looks good, 
not sure if we want to also match uppercase A-F in which case you could use `[0-9a-fA-F]` but if not, this commit looks good

Issue #187, opened 2022-02-14
Remove Summary and Configuration


rmarx:
We currently have the `summary` (in QlogFile) and `configuration` (in Trace) fields.

These fields made sense when qlog was viewed as a file format that would primarily be used to aggregate multiple traces into a single file, which would then be transferable between different tools (e.g., in research settings). 

However, with most implementations shifting to streaming logging (1 trace per file) and no tools emerging that rely on interoperability via the `summary` and `configuration` fields, I feel it makes sense to just remove them from the spec. This is further motivated by the fact that Summary is defined as just a generic JSON object and Configuration only has 2 recommended fields, 1 of which is `original_uris`, which again only makes sense if we see this as a grouping format. 

See also #184, cc @LPardue

LPardue:
+1 to simplicity. 

It seems like it's going to be straightforward for anyone to add such fields in future if they really need them. Being judicious with we things we include in the base spec will make it more straightforward for people to focus on the valuable items.

PR #208, opened 2022-06-29
add an environment field to the trace

### LINKED ISSUE #136: add a field to log the code version
With many iterations of my qlog implementation running in production now, it would be helpful if there was a way to tell which version of my code generated any particular qlog file (even better if qvis could display it somewhere).

I suggest adding an optional `code_version` to the `Trace` container. Implementations could fill this field with whatever they want. In quic-go, I'd probably use something like `quic-go $(git describe --long --dirty)`.

marten-seemann:
Fixes #136.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -377,6 +377,7 @@ Trace = {
     ? title: text
     ? description: text
     ? configuration: Configuration
+    ? environment: text
     ? common_fields: CommonFields
     ? vantage_point: VantagePoint
     events: [* Event]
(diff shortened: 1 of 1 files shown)


lnicco (commented):
As @rmarx recommended on #136: could the `configuration` field above serve the same purpose? 

From the current text:
```
The configuration field can be viewed as a generic metadata field that tools can
fill with their own fields, based on per-tool logic. 
```

LPardue:
I suggest a hybrid - put the environment field in the Configuration object.

LPardue:
Although, https://github.com/quicwg/qlog/issues/187 suggests removing the configuration field. So we have a conflict of idealogies going on.,

rmarx (changes requested):
See comment on #136

PR #209, opened 2022-06-29
introduce a type for stateless reset tokens

### LINKED ISSUE #122: the new Token is confusing

marten-seemann:
Fixes #122.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1519,7 +1519,7 @@ PacketHeader = {
 
 ~~~ cddl
 Token = {
-    ? type: "retry" / "resumption" / "stateless_reset"
+    ? type: "retry" / "resumption"
 
     ; byte length of the token
     ? length: uint32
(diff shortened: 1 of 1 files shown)


lnicco (commented):
if not using the `Token` type for stateless reset anymore you may want to remove `"stateless_reset" as the Token type here https://fburl.com/7t5zaup7

marten-seemann:
Thanks! Fixed.

LPardue (approved):

PR #211, opened 2022-07-18
Add optional src and dst address for `TransportPacket*` events

### LINKED ISSUE #57: Add IP addresses to datagram_* events
We currently extracted these to other events, assuming they wouldn't change or only change sparingly at specific times, handled by other events (e.g., when doing migration).

However, this is not always the case. For example, @huitema mentioned:

> NAT rebinding, probes of packets before migration

For these cases, it would be interesting to have (optional) IP from/to addresses in the datagram_* events as well.

lnicco:
Add optional src and dst address for TransportPacketSent, TransportPacketReceived, TransportPacketDropped, TransportPacketBuffered

Fixes #57

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -778,6 +778,8 @@ TransportPacketSent = {
 
     ? raw: RawInfo
     ? datagram_id: uint32
+    ? src_ip: IPAddress
+    ? dst_ip: IPAddress
 
     ? trigger:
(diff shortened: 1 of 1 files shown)


lnicco:
Actually this should be for the Datagram events and not for Packet events 
TransportDatagramsSent
TransportDatagramsReceived
TransportDatagramDropped

LPardue:
Overtaken by #336?

rmarx:
This was indeed solved (or at least changed in a different way) by #336, so closing without merging.

PR #213, opened 2022-09-08
give some advice how to associate VN and Retry with connections

### LINKED ISSUE #105: Log Retry and Connection in same Qlog file, use ODCID for indentification
Many implementations identify the Qlog file using the "Initial DCID" used by the client. In the case of Retry, this results in two Qlog traces: one identified by the original DCID that contains the first initial packet of the client and the Retry packet that the server produced, and a second one identified by the Initial DCID used in the subsequent connection.

My proposal is to use a single file for the retry and for the next connection, identified by the ODCID. But I wonder whether that's going to break something.

marten-seemann:
Fixes #105.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -245,21 +245,15 @@ events.
 Ideally, implementers SHOULD create a separate, individual "endpoint-level" trace
 file (or group_id value), not associated with a specific connection (for example a
 "server.qlog" or group_id = "client"), and log all events that do not belong to a
-single connection to this grouping trace. However, this is not always practical,
-depending on the implementation. Because the semantics of most of these events are
-well-defined in the protocols and because they are difficult to mis-interpret as
-belonging to a connection, implementers MAY choose to log events not belonging to
-a particular connection in any other trace, even those strongly associated with a
-single connection.
+single connection to this grouping trace.
 
-Note that this can make it difficult to match logs from different vantage points
-with each other. For example, from the client side, it is easy to log connections
-with version negotiation or retry in the same trace, while on the server they
-would most likely be logged in separate traces. Servers can take extra efforts
-(and keep additional state) to keep these events combined in a single trace
-however (for example by also matching connections on their four-tuple instead of
-just the connection ID).
-
+There are events that are conceptually associated with a single connection, but
+that happen before the actual QUIC connection is established when viewed from
+the server's perspective. This includes version negotiation as well as retries.
+Servers MAY take extra efforts to associate these events with a successfully
+established QUIC connection and combine them in a single trace. For version
+negotiation, matching can be done by identifying a connection attempt by its
+four-tuple, and retries can be identified by their ODCID.
 
 
 # QUIC event definitions


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -245,21 +245,15 @@ events.
...
-
+single connection to this grouping trace.
+
+There are events that are conceptually associated with a single connection, but
+that happen before the actual QUIC connection is established when viewed from
+the server's perspective. This includes version negotiation as well as retries.
+Servers MAY take extra efforts to associate these events with a successfully
+established QUIC connection and combine them in a single trace. For version
+negotiation, matching can be done by identifying a connection attempt by its
+four-tuple, and retries can be identified by their ODCID.
```
LPardue, line 256: I don't quite  like the way the text read says "MAY take extra effort". Mainly because this text is also there to be read by consumers who want to know what events to expect but don't care about effort. 

A suggested rewording to avoid. Other options would work just too.

```suggestion
Servers MAY include such events in the connection trace but need to ensure those
events are correctly associated with the correct connection, which can take
extra effort or resource. For example, version negotiation events can be matched
to a connection by identifying a connection attempt by its four-tuple;
similarly, retry events be matched by their ODCID.
```

marten-seemann: Suggestion sounds good. Thank you!

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -245,21 +245,16 @@ events.
...
+single connection to this grouping trace.
+
+There are events that are conceptually associated with a single connection, but
+that happen before the actual QUIC connection is established when viewed from
+the server's perspective. This includes version negotiation as well as retries.
+Servers MAY include such events in the connection trace but need to ensure those
+events are correctly associated with the correct connection, which can take
+extra effort or resource. For example, version negotiation events can be matched
+to a connection by identifying a connection attempt by its four-tuple;
+similarly, retry events be matched by their ODCID.
```
LPardue, line 257: think this was my fault, sorry

```suggestion
similarly, retry events can be matched by their ODCID.
```

LPardue:
@huitema could you take a look at this PR and see if it addresses your issue in #105 please.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -245,21 +245,16 @@ events.
...
+single connection to this grouping trace.
+
+There are events that are conceptually associated with a single connection, but
+that happen before the actual QUIC connection is established when viewed from
+the server's perspective. This includes version negotiation as well as retries.
+Servers MAY include such events in the connection trace but need to ensure those
+events are correctly associated with the correct connection, which can take
+extra effort or resource. For example, version negotiation events can be matched
+to a connection by identifying a connection attempt by its four-tuple;
+similarly, retry events can be matched by their ODCID.
```
huitema, line 258: In a server farm, there is no guarantee that VN and later connection are handled by the same server -- in contrast with the ODCID mechanism that can be used to achieve that. That means it could be hard to associate VN and connection at logging time. And then, the is also no guarantee that either Retry or VN is followed by an actual connection. Retry may be caused by a DDOS attack. VN can be caused by something like [quicping](https://github.com/ooni/probe/issues/1994). In fact, I could see a need for an easy way to track orphans, and see how the world is using or abusing the server. Something like a global log, with events like VN, Retry, Connection arrival, Connection final.

LPardue: This makes sense. I think it is already covered by the preceeding paragraph though? Maybe what we need to say is that events can be put in multiple logs - the global one, and the conenction-specific one if the match succeeds.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -245,21 +245,16 @@ events.
...
-just the connection ID).
-
+single connection to this grouping trace.
+
+There are events that are conceptually associated with a single connection, but
+that happen before the actual QUIC connection is established when viewed from
+the server's perspective. This includes version negotiation as well as retries.
+Servers MAY include such events in the connection trace but need to ensure those
+events are correctly associated with the correct connection, which can take
+extra effort or resource. For example, version negotiation events can be matched
```
marten-seemann, line 255:
```suggestion
extra effort or resources. For example, version negotiation events can be matched
```

marten-seemann:
I would claim that PR is obsolete, now that we have https://github.com/quicwg/qlog/pull/279. Unless anybody objects within the next 24h, I'm going to close this PR and the associated issue (#105).

PR #217, opened 2022-09-08
Improve referencing

### LINKED ISSUE #216: Use QUIC and H3 RFCS
qlog alreaady refers to the -34 drafts. Nominally this change should be easy. Although some section references might need double checking.

LPardue:
First, these documents can now refer to the published QUIC RFCs.

While doing that, let's just replace all [REF] with {{REF}}, which
works better for our tooling.

Lastly, lets brush up the cross references between documents a bit.

Closes #216

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -32,37 +32,8 @@ author:
 
 normative:
 
-  QUIC-HTTP:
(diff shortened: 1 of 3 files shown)


marten-seemann (approved):
lgtm, assuming that the build tool correctly sets the links to the RFCs

LPardue:
I checked this locally and everything was good. It's easy enough to fix up later if I did break something.

PR #224, opened 2022-09-09
Define HTTPUnknownFrame type for H3

### LINKED ISSUE #201: UnknownFrame in QUIC and HTTP/3 events.
In https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html#appendix-A.3.9 you say

> HTTP/3 qlog re-uses QUIC's UnknownFrame definition, since their values and usage overlaps

This is ambiguous to me. Are you saying the defined object - QUIC `UnknownFrame` - is reused, or that the structure of the object - HTTP/3 `UnknownFrame - is identical.

I think it has to be the latter, because some HTTP/3 events carry `HTTPFrame`. If this is the case, I would just fully define the frame in the HTTP/3 specification to avoid confusion.

LPardue:
Avoid confusing use of QUIC's UnknownFrame type.

Fixes #201

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -599,7 +599,7 @@ HTTPFrame =  HTTPDataFrame /
              HTTPGoawayFrame /
              HTTPMaxPushIDFrame /
              HTTPReservedFrame /
-             UnknownFrame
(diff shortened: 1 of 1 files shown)


lnicco (approved):
Thanks!

PR #226, opened 2022-09-12
Flatten event definition ToC, add summary table

LPardue:
This is a subjective editorial change. The intent is to improve the
readability of the document and expose better how we expect people
to use the "name" field. By splitting out the overview of events
from the definitions themselves, we achieve those intents.


LPardue:
Easier to review as 

[HTML](https://quicwg.org/qlog/summary-table)
[diff](https://www.ietf.org/rfcdiff?url1=https://quicwg.github.io/qlog/draft-ietf-quic-qlog-main-schema.txt&url2=https://quicwg.github.io/qlog/summary-table/draft-ietf-quic-qlog-main-schema.txt)

LPardue:
I have no issue with switching to consistent names. 

Initially I wanted to make the table multi-column and to keep its identifiers short. That's not how it turned out, so I'll push a fixup.

LPardue:
> Finally, this should have accompanying and similar PRs for the main and H3 documents down the line (maybe make an issue so we don't forget those?)

Yes. I didn't want to invest any more effort if the editor team didn't like this approach. I'd suggest we refine this PR, get it merged, and then update the other documents using this as a template.

PR #227, opened 2022-09-12
Add unknown HTTPStreamType

### LINKED ISSUE #205: HTTPStreamType needs an unknown enumeration
It's possible that a peer would open stream types of unknown type, and not in the reserved range. So let's fix that up.

LPardue:
Closes #205

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -249,6 +249,9 @@ HTTPStreamTypeSet = {
     ? old: HTTPStreamType
     new: HTTPStreamType
 
+    ; only when new === "unknown"
+    ? raw_stream_type: uint64
+
     ; only when new === "push"
     ? associated_push_id: uint64
 }
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -257,6 +257,7 @@ HTTPStreamType =  "data" /
                   "control" /
                   "push" /
                   "reserved" /
+                  "unknown" /
                   "qpack_encode" /
                   "qpack_decode"
 ~~~


rmarx:
I wonder if this needs some way to set the raw stream type instead of just logging "unknown"

We have precedent for this in the quic-events with:

```
ConnectionCloseFrame = {
    frame_type: "connection_close"

    ? error_space: ErrorSpace
    ? error_code: TransportError / $ApplicationError / uint32
    ? raw_error_code: uint32
    ? reason: text

    ; For known frame types, the appropriate "frame_type" string
    ; For unknown frame types, the hex encoded identifier value
    ? trigger_frame_type: uint64 / text
}
```

and 

```
UnknownFrame = {
    frame_type: "unknown"
    raw_frame_type: uint64

    ? raw_length: uint32
    ? raw: hexstring
}
```

LPardue:
Looging the unknown value seems like a good idea. But there's no frame parsing going on here, it's just the first varint on the stream. So I think we'd need instead

```
HTTPStreamTypeSet = {
    ... all the existing stuff
    
    ; add 
    raw_stream_type: uint64
``

WDYT?

**Comment by rmarx** [comment]
I seem to have been conflating StreamType with FrameType here in my suggestion... yours is of course much better.

If included in StreamTypeSet though, I'd make it an optional field (as the known types don't have to set it), so

```
HTTPStreamTypeSet = {
    ... all the existing stuff
    
    ; This should only be present if the value of new is "unknown" 
    ? raw_stream_type: uint64
```

LPardue:
oops I meant it to be optional, and failed. I'll update the PR a bit later to reflect your suggestion. Cheers.

marten-seemann (commented):
Sorry if that's a stupid question, but why do we have an `old` and a `new` field here? I thought you can only set it once.

LPardue:
Seems a valid question. Spun off to #236

PR #231, opened 2022-09-12
Notational conventions aren't the place to discuss

LPardue:
There's too much text here that doesn't aid the reader. Delete it to be
clear that we're using JSON as the notational convention for examples.

If this text was valuable, it should be somewhere else.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -184,15 +184,8 @@ Other qlog documents can define their own CDDL-compatible (struct) types
 
 ### Serialization
 
-While the qlog schemas are format-agnostic, and can be serialized in
-many ways (e.g., JSON, CBOR, protobuf, ...), this document only
-describes how to employ {{!JSON=RFC8259}}, its subset
-{{!I-JSON=RFC7493}}, and its streamable derivative
-{{!JSON-Text-Sequences=RFC7464}} as textual serialization options. As
-such, examples are provided in {{!JSON=RFC8259}}. Other documents may
-describe how to utilize other concrete serialization options, though
-tips and requirements for these are also listed in this document
-({{concrete-formats}}).
+The qlog schema can be serialized in many ways (e.g., JSON, CBOR, protobuf,
+etc). Serialization examples in this document use JSON ({{!JSON=RFC8259}}).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -184,15 +184,8 @@ Other qlog documents can define their own CDDL-compatible (struct) types
...
-many ways (e.g., JSON, CBOR, protobuf, ...), this document only
-describes how to employ {{!JSON=RFC8259}}, its subset
-{{!I-JSON=RFC7493}}, and its streamable derivative
-{{!JSON-Text-Sequences=RFC7464}} as textual serialization options. As
-such, examples are provided in {{!JSON=RFC8259}}. Other documents may
-describe how to utilize other concrete serialization options, though
-tips and requirements for these are also listed in this document
-({{concrete-formats}}).
+The qlog schema can be serialized in many ways (e.g., JSON, CBOR, protobuf,
+etc). Serialization examples in this document use JSON ({{!JSON=RFC8259}}).
```
LPardue, line 188: We'll keep the definition of serialization example using JSON. But we need to keep the talk about serialization considerations somewhere else

rmarx:
I still feel we should have a "forward reference" here, indicating that we also discuss I-JSON and Streaming JSON considerations in this document, rather than making it seem we only do "pure JSON" due to the examples.

Put differently, something like this should still be mentioned in this location imo:

```
This document describes how to employ {{!JSON=RFC8259}}, its subset
{{!I-JSON=RFC7493}}, and its streamable derivative
{{!JSON-Text-Sequences=RFC7464}}.
```

LPardue:
But the fact it can be serialized to I-JSON has nothing to do with notational conventions used in this document, because we don't use any I-JSON.

rmarx:
@LPardue I just pushed a commit with a proposal to split up the original text to keep only a minimal amount in "Notational Conventions" but keep the core message alive in the introduction (which seemed the most logical location for now). 

I added the "unless otherwise indicated", as there are some examples of JSON-SEQ near the end of the document that obviously don't use RFC8259 notation directly. 

Please review and if approved, merge :)

rmarx (approved):

marten-seemann (approved):

PR #235, opened 2022-09-16
Update MTU event definitions

rmarx:
Updates #210 as that had some problems:

- used `boolean` instead of `bool` for CDDL type in some of the definitions (not in others)...
- Made a separate category for the `mtu_updated` event. It's debatable imo whether it belongs in `connectivity` or `transport`, but it shouldn't be its own thing. 

In general, I feel #210 was approved/merged too soon and this type of things should be checked for future PRs of a similar type.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -426,15 +426,16 @@ For now, infer from other connectivity events and path_challenge/path_response f
 Importance: Extra
 
 ~~~ ccdl
-MTUUpdated = {
-  ? mtu: uint16
+ConnectivityMTUUpdated = {
+  ? old: uint16
+  new: uint16
 
   ; at some point, MTU discovery stops, as a "good enough"
   ; packet size has been found
   ? done: bool .default false
 }
 ~~~
-{: #mtu-updated-def title="MTUUpdated definition"}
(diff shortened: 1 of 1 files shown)


marten-seemann (approved):

marten-seemann:
Changes look good to me.

> In general, I feel https://github.com/quicwg/qlog/pull/210 was approved/merged too soon and this type of things should be checked for future PRs of a similar type.

First of all, sorry for messing up the type name. I'm not sure I agree that it was merge too soon though, #210 was open for 3 months, and was approved by one of the editors. Sometimes it's better to make progress instead of not moving at all, even if some cleanup is necessary later.

rmarx:
@marten-seemann no problem, I mess up way worse than that :)

Let me rephrase that to just "approved" too soon then, instead of "merged" too soon (which is again perfectly understandable with Lucas not used to watching for that type of issue, just wanted to indicate future approvals should).

rmarx:
From the discussion in #236, it seems better to keep this consistent with other `*Updated` events.

Also noticed I forgot to prefix the event with Connectivity and add it to the CDDL list on the bottom...

LPardue (approved):

LPardue (approved):

marten-seemann (approved):

PR #244, opened 2022-09-29
Make packet_number optional

### LINKED ISSUE #133: Clarify that Retry, VNEG, and Stateless Reset packets don't have packet numbers
With draft-02, we moved packet_number to PacketHeader and used that field for all types of packets as well. 
However, packet_number is currently a required field, but Retry, Version Negotiation and Stateless Reset packets don't actually use packet numbers.

I don't want to just make the packet_number field generally optional, since imo it shouldn't be for other packet types. Better (best?) solution is to add an explicit comment indicating the field MUST be omitted for the 3 exceptions and MUST be included for all others.

rmarx:
Fixes #133

(not sure whether to indent with spaces to vertically match the listing or to just let it wrap to the next line without prepended spaces though)

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1422,7 +1422,7 @@ PacketNumberSpace = "initial" / "handshake" / "application_data"
 ~~~ cddl
 PacketHeader = {
     packet_type: PacketType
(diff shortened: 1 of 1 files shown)


rmarx (commented):
remove trailing space

LPardue (approved):

Issue #221, opened 2022-09-09
Putting event data definitions in the appendix doesn't feel right


LPardue:
This also relates to https://github.com/quicwg/qlog/issues/220

In both the QUIC and H3 documents, the interesting data of each event is defined in the appendix. Typically, appendicesare used for only additional or informative content. 

Since the data is part of the normative schema, I'd suggest the definitions need to be promoted into the main body of the text.

rmarx:
Agreed this shouldn't be in appendix.

Do want to keep the data definitions together though, instead of e.g., defining each where they are first used in an event. 

Since you are splitting the event categories into a Section per Category @LPardue, would this mean 1 new top-level Section for the data definitions (so just lift them up from appendices as they are) or split out more?

PR #249, opened 2022-10-15
Extract H3 data definitions out of the appendix

LPardue:
Part of addressing #221. I started with the h3 doc because it's smaller and simpler. This moves the category specific data definitions to be underneath the respective HTTP/3 or QPACK section of the document. 

The markdown diff is hard to assess, better to look at the html and diff preview

* https://quicwg.org/qlog/move-h3-data-outta-appendix/draft-ietf-quic-qlog-h3-events.html
* https://www.ietf.org/rfcdiff?url1=https://quicwg.github.io/qlog/draft-ietf-quic-qlog-h3-events.txt&url2=https://quicwg.github.io/qlog/move-h3-data-outta-appendix/draft-ietf-quic-qlog-h3-events.txt

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -365,18 +365,18 @@ HTTPPushDecision = "claimed" / "abandoned"
 ~~~
 {: #http-pushresolved-def title="HTTPPushResolved definition"}
 
-## HTTP/3 Data Field Definitions
+# HTTP/3 Data Field Definitions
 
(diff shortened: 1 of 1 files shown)


LPardue:
Note that RFC format doesn't like more than 3 levels in a table of contents. This means that specific h3 frames, qpack instructions and block prefixes are not shown. If we think that matters, we either need to promote the data defitions one level, or flatten them into the thrid-level.

rmarx (approved):

PR #253, opened 2022-10-19
H3 CDDL: use consistent spacing

lnicco:
in the H3 CDDL definitions we have been using extra lines to space things out, but it is a bit inconsistent. 
This PR tries to make things more consistent by
  * only keep empty lines between types in the same block
  * split unions into multiple lines, except for when only two types are present

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -191,9 +191,7 @@ Definition:
 ~~~ cddl
 HTTPParametersSet = {
     ? owner: Owner
-
     ~HTTPParameters
-
     ; qlog-specific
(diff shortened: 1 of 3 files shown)


lnicco:
Discussed in the meeting. Some TODOs:
 * one empty line before each comment line (but not after)
 * always use multiple lines for union types: even if there are just two alternatives
 * apply the changes to the other docs as well

LPardue (approved):

lnicco:
rebasing was a nightmare so I created a new (draft) PR https://github.com/quicwg/qlog/pull/298  

I can force push the new commits here in case that is better. Let me know

lnicco:
fixed in #298

PR #255, opened 2022-10-19
Remove connection_retried event placeholder

rmarx:
Connection retried was never properly specced. 

The main reason is that a retried connection is completely separate from the original connection attempt, except for the `retry token` value, and as such is usually logged in a separate file / separate context from the original connection qlog data. 

Current qlog deployments either have other ways of correlating retried connections with the original attempts, or don't need to do this. Either way, there is no tangible need for `connection_retried` at this time.


marten-seemann:
> The main reason is that a retried connection is completely separate from the original connection attempt, except for the retry token value, and as such is usually logged in a separate file / separate context from the original connection qlog data.

I don't think that's correct. You also take your Connection IDs with you to the new connection. quic-go logs it to the same file, as _it is the same connection attempt_.

I don't see value in `connection_retried` though.

marten-seemann (approved):

LPardue (approved):

PR #258, opened 2022-10-24
Take a stab at refactoring the use of `Note:`.

lnicco:
This is just a start. While these are editorial changes they will
definitely require discussion for how it's best to change each Note:.


LPardue (approved):
LGTM!

PR #263, opened 2022-12-28
add ECN events

### LINKED ISSUE #212: add ECN events
Endpoints using ECN will perform validation of ECN, as described in https://datatracker.ietf.org/doc/html/rfc9000#section-13.4.2. There are a few different conditions that can lead to a failure of ECN validation, which in turns leads to disabling of ECN for that path. We should have an event that tells us _that_ ECN validation failed and _why_.

marten-seemann:
Fixes #212.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -699,8 +699,6 @@ QUICPacketSent = {
     ? datagram_id: uint32
     ? is_mtu_probe_packet: bool .default false
 
-    ? ecn: ECN
-
     ? trigger:
       ; draft-23 5.1.1
       "retransmit_reordered" /
@@ -746,8 +744,6 @@ QUICPacketReceived = {
     ? supported_versions: [+ QuicVersion]
     ? raw: RawInfo
     ? datagram_id: uint32
-
-    ? ecn: ECN
 
     ? trigger:
         ; if packet was buffered because it couldn't be
@@ -881,6 +877,11 @@ QUICDatagramsSent = {
     ; The RawInfo fields do not include the UDP headers,
     ; only the UDP payload
     ? raw: [+ RawInfo]
+
+    ; ECN bits in the IP header
+    ; if not set, defaults to Non-ECT
+    ? ecn: [+ ECN]
+
     ? datagram_ids: [+ uint32]
 }
 ~~~
@@ -914,6 +915,11 @@ QUICDatagramsReceived = {
     ; The RawInfo fields do not include the UDP headers,
     ; only the UDP payload
     ? raw: [+ RawInfo]
+
+    ; ECN bits in the IP header
+    ; if not set, defaults to Non-ECT
+    ? ecn: [+ ECN]
+
     ? datagram_ids: [+ uint32]
 }
 ~~~
@@ -1674,7 +1680,7 @@ AckFrame = {
     ; ECN (explicit congestion notification) related fields
     ; (not always present)
     ? ect1: uint64
-    ? ect0:uint64
+    ? ect0: uint64
     ? ce: uint64
 
     ; total frame length, including frame header
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1569,7 +1569,7 @@ KeyType = "server_initial_secret" /
 
 ## ECN
 ~~~ cddl
-ECN = "00" / "01" / "10" / "11"
+ECN = "Non-ECT" / "ECT(1)" / "ECT(0)" / "CE"
 ~~~
 {: #ecn-def title="ECN definition"}
 
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -880,7 +880,8 @@ QUICDatagramsSent = {
     ? raw: [+ RawInfo]
 
     ; ECN bits in the IP header
-    ; if not set, defaults to Non-ECT
+    ; if not set, defaults the value used on the last
+    ; QUICDatagramsSent event
     ? ecn: [+ ECN]
 
     ? datagram_ids: [+ uint32]
@@ -918,7 +919,8 @@ QUICDatagramsReceived = {
     ? raw: [+ RawInfo]
 
     ; ECN bits in the IP header
-    ; if not set, defaults to Non-ECT
+    ; if not set, defaults to the value on the last
+    ; QUICDatagramsReceived event
     ? ecn: [+ ECN]
 
     ? datagram_ids: [+ uint32]
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -880,7 +880,7 @@ QUICDatagramsSent = {
     ? raw: [+ RawInfo]
 
     ; ECN bits in the IP header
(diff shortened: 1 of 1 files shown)


marten-seemann:
2023-01-05 conversation:
* should this be on the datagram_ instead of packet_ event?
* is there a default value? Should this be a separate event like the spin bit?
* use semantic names?

marten-seemann:
I've updated the PR to:
* use the semantic names instead of the two bits
* defined Non-ECT as the default value
* moved this to the datagram_ events

While technically more correct, I'm not a big fan of the last one, as the datagram_ events have an importance of "Extra", so it's more likely ECN won't be logged. For example, quic-go currently doesn't use the datagram_ events at all.

rmarx:
I think that's a fair point about `datagrams_` vs `packet_`. IIUC, the only place where this matters anyway is during the handshake, as there can be only a single 1-RTT packet in one datagram. I'm not sure if/how much ECN comes into play during the handshake?

I think the argument could be made that for the handshake, the ECN bits could be reflected on/copied to each of the potentially coalesced QUIC `packet_` events. They can still be tracked to the singular datagram they were in using the `datagram_id` construction after all. 

It's a bit less "clean"/"correct", but that can be said of many things in qlog ;) What do you think, @LPardue?

marten-seemann:
We could also put it on both event types, and allow implementations that don’t log datagram events to put it in the packet. Of course, this shifts some complexity to consumers of the qlog.

marten-seemann:
I added events for the ECN testing state machine from RFC 9000, A.4. This should now cover the (non-L4S) use case of ECN.

Would appreciate a review of this PR.

rmarx:
This looks good to me overall, but I wonder if it would be better to make the ECN enum extensible as well (to allow for easy expansion into L4S in the future if needed):

e.g., something like

```
$ECNfield /= {
    * text => any
}

ECN = "Non-ECT" / "ECT(1)" / "ECT(0)" / "CE"

$ECNfield /= ECN
```

Additionally, I wonder if we want some way to remove some chattiness here... I imagine that most values in DatagramsSent will be identical once the probing is done, and that you really only want to log when a CE is received in DatagramsReceived. Not sure if it's needed, but maybe some text to state that explicitly like "no need to log with EVERY packet, only if values change from previously logged value"? (though arguably you could say that's also observable through the reflected values in the ACKs...).

marten-seemann:
> This looks good to me overall, but I wonder if it would be better to make the ECN enum extensible as well (to allow for easy expansion into L4S in the future if needed):
> 
> e.g., something like
> 
> ```
> $ECNfield /= {
>     * text => any
> }
> 
> ECN = "Non-ECT" / "ECT(1)" / "ECT(0)" / "CE"
> 
> $ECNfield /= ECN
> ```

I don't think that's needed. There are only 4 ECN values (2 bits) allowed by the IP header. L4S doesn't change that.



> Additionally, I wonder if we want some way to remove some chattiness here... I imagine that most values in DatagramsSent will be identical once the probing is done, and that you really only want to log when a CE is received in DatagramsReceived. Not sure if it's needed, but maybe some text to state that explicitly like "no need to log with EVERY packet, only if values change from previously logged value"? (though arguably you could say that's also observable through the reflected values in the ACKs...).

It's not observable from the ACKs, the ACK only tells you what happened along the path. Not logging it on every packet sounds good though.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1561,6 +1599,14 @@ KeyType = "server_initial_secret" /
 ~~~
 {: #keytype-def title="KeyType definition"}
 
+## ECN
+~~~ cddl
+ECN = "Non-ECT" / "ECT(1)" / "ECT(0)" / "CE"
```
marten-seemann, line 1604: See https://en.wikipedia.org/wiki/Explicit_Congestion_Notification#Operation_of_ECN_with_IP
```suggestion
ECN = "Not-ECT" / "ECT(1)" / "ECT(0)" / "CE"
```

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1413,6 +1426,31 @@ RecoveryMarkedForRetransmit = {
...
+ECNStateUpdated = {
+   ? old: ECNState
+    new: ECNState
+}
+
+ECNState =
+  ; ECN testing not yet started, ECN support not determined yet.
+  "unknown" /
+  ; ECN testing in progress
+  "testing" /
```
marten-seemann, line 1445: This is backwards. Testing comes first, and after that comes unknown.
```suggestion
  ; ECN testing in progress
  "testing" /
  ; ECN state unknown, waiting for acknowledgements for testing packets
  "unknown" /
```

rmarx:
> It's not observable from the ACKs, the ACK only tells you what happened along the path.

not sure what you mean by this. The ACKs reflect the incoming ECN info to the sender, so looking at incoming ACKs shows you what happened from YOU -> THEM and looking at outgoing ACKs (the one's we're sending) should show what happened THEM -> US. Put differently: your outgoing ACKs should reflect what you saw in DatagramsReceived, no? (though, to be fair, maybe I'm misunderstanding how all of this works).

marten-seemann:
> > It's not observable from the ACKs, the ACK only tells you what happened along the path.
> 
> not sure what you mean by this. The ACKs reflect the incoming ECN info to the sender, so looking at incoming ACKs shows you what happened from YOU -> THEM and looking at outgoing ACKs (the one's we're sending) should show what happened THEM -> US. Put differently: your outgoing ACKs should reflect what you saw in DatagramsReceived, no? (though, to be fair, maybe I'm misunderstanding how all of this works).

You're right, for the receive side. For the send side this doesn't work though.

rmarx:
Ok, agreed then :) 
Though, as I said, for the send side the codepoints will (probably?) usually always be the same after detection/negotiation/testing (e.g., if capable, you always send ECT(0) from then on), so no need to keep logging after that. Hence the original comment above.

marten-seemann:
That makes sense. I updated the definition such that the default is now the value on the last event of that type (c7cd9bbcf21e7065ef51c26cb726b5c870522221).

@LPardue @rmarx Would you mind taking another look at this PR? Would love to get it in before the draft cutoff date for Prague.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -875,6 +878,12 @@ QUICDatagramsSent = {
     ; The RawInfo fields do not include the UDP headers,
     ; only the UDP payload
     ? raw: [+ RawInfo]
+
+    ; ECN bits in the IP header
+    ; if not set, defaults the value used on the last
```
LPardue, line 883:
```suggestion
    ; if not set, defaults to the value used on the last
```

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1413,6 +1428,31 @@ RecoveryMarkedForRetransmit = {
 ~~~
 {: #recovery-markedforretransmit-def title="RecoveryMarkedForRetransmit definition"}
 
+## ecn_state_updated {#recovery-ecnstateupdated}
+Importance: Extra
+
+This event indicates a progression in the ECN state machine as described in section
+A.4 of RFC 9000.
```
LPardue, line 1435:
```suggestion
A.4 of {{QUIC-TRANSPORT}}.
```

LPardue (approved):
LGTM modulo 2 nits

PR #264, opened 2022-12-28
don't call the AEAD auth tag "trailer"

marten-seemann:
Neither RFC 9000 nor 9001 call this a "trailer". We should avoid this terminology, since it could potentially be confused with HTTP trailers.


rmarx:
First make a decision on #265, then revise this one, as it might need changes depending on that.

PR #265, opened 2022-12-28
remove the aead_tag_length field from TransportParametersSet

marten-seemann:
This field is not particularly useful, because
1. it will always be 16, see https://datatracker.ietf.org/doc/html/rfc9001#name-aead-usage
2. it's a property of the TLS cipher suite in use, so it can be trivially looked up, if necessary

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -108,7 +108,7 @@ trace of the connection with ODCID abcd1234).
 
 QUIC packets always include an AEAD authentication tag at the end.
 In general, the length of the AEAD tag depends on the TLS cipher
-suite, although all cipher suites used for QUIC use a 16 byte tag.
+suite, although all cipher suites used in QUIC v1 use a 16 byte tag.
 
 Note:
 
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -106,20 +106,16 @@ trace of the connection with ODCID abcd1234).
 
 ## Raw packet and frame information
 
-Note:
-
-: QUIC packets always include an AEAD authentication tag at the end.
-As this tag is always the same size for a given connection (it depends on the
-used TLS cipher suite), this document does not define a separate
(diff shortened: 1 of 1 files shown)


lnicco (approved):
looks good to me

rmarx:
So this is a bit of a tough one for me, since I've gone through the pain of trying to determine the length of the TLS trailer for TLS-over-TCP from wireshark output that does NOT explicitly include the size and also makes it difficult to get the proper cipher name to heuristically determine it. 

That said, for QUIC V1 it's indeed all 16, so you -could- make an argument we can drop it here and add it back in with a V2.1/V3 if new ciphers start being used. I don't have a particularly big preference either way.

If we do this change though, the prose could use tightening, because in the current version we have 2 sentences after another that indicate the aead_length depends on the cipher suite.

LPardue:
I'm leaning on the side of removing this field. The information is made redundant by the cipher suite. Things that care about the tag length can do a lookup or make their own assumptions based on other information

rmarx:
Discussed on call. @marten-seemann will make the requested prose change and should be good to go.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -106,20 +106,16 @@ trace of the connection with ODCID abcd1234).
...
-Note:
-
-: QUIC packets always include an AEAD authentication tag at the end.
-As this tag is always the same size for a given connection (it depends on the
-used TLS cipher suite), this document does not define a separate
-"RawInfo:aead_tag_length" field here. Instead, this field is reflected in
-"transport:parameters_set" and can be logged only once.
+QUIC packets always include an AEAD authentication tag at the end.
+In general, the length of the AEAD tag depends on the TLS cipher
+suite, although all cipher suites used for QUIC use a 16 byte tag.
```
rmarx, line 111: I wonder if we should be explicit in saying "used for QUIC v1" or "used in RFC9001" for future-proofness?

rmarx:
Minus the comment above, this LGTM.

rmarx (approved):

rmarx (approved):

Issue #72, opened 2020-03-28
Allow listing which events were supported by the logger


rmarx:
When looking at a qlog at this time, if there is a certain event type missing, you don't know if it's because those events never occurred or if the implementation simply doesn't support/log that event type. 

This is especially troublesome if you want to selectively enable/disable certain types to reduce overhead when you're doing targeted "live" debugging on a deployed system.

The simplest solution I can see is to add a new field to `configuration`, for example:

> supportedEvents:Array\<string\> // each string is category:event_type (e.g., transport:packet_sent)

Some bikeshedding can help to decide if we need two separate fields (supportedCategories separately for example) or if we can cut out the category in here completely (as there is little overlap in event names atm).

This would also be interesting for doing a fast "is this trace compatible with this visualization"-check in qvis and others tools. 

Thanks to @mjoras for reporting.

marten-seemann:
This only captures a subset of the possible changes. For example, you wouldn't be able to tell if a  qlog implementation already supported the recently added `stateless_reset` packet type or not.

In the end, I think it makes more sense to add some kind of versioning scheme to your qlog file (which is orthogonal to the `qlog_version`), and would not be interpreted by tools.

rmarx:
I'm not sure I follow you here...
If the implementation doesn't support stateless_reset, it won't log them either in the first place, so them being absent from the `supportedEvents` list is the same.

This is not really intended to show what an implementation "supports" but more what is has "enabled". The situation is that as a user, you get a random qlog file in front of you, on which you need to do analysis. You're looking for the occurrence of a certain event type, but it isn't there. Is that because the event type wasn't being logged for this log (e.g., mvfst wants to selectively log only -some- of the supported event types to keep overhead down) or because it actually never occurred during the connection (e.g., signalling the bug you're looking for).

Could you expand on the orthogonal versioning scheme? For your example of the `stateless_reset`, that would just be `draft-02` going forward from my viewpoint.

LPardue:
I think this can be resolved as 2 optional lists containing event category+type, or name. One list  declares what events were in scope to be logged, one that declares what events were actually logged.

This would allow for easier scanning of traces to determine if they have events of interest - I.e. speeding up finding 1 in 10,000 connections that fail or bwhave in a certain way

LPardue:
[...]

PR #269, opened 2023-01-02
can a hexstring field be empty ?

lnicco:
A zero-length connection id is the only use-cases I can find where an empty hexstring would make sense. But we are using an optional field for those?
This is a minor thing. Posting for discussion more than anything

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -125,7 +125,7 @@ uint64 = uint .size 8
 ; examples: 82dc, 027339, 4cdbfd9bf0
 ; this is needed because the default CDDL binary string (bytes/bstr)
 ; is only CBOR and not JSON compatible
-hexstring = text .regexp "([0-9a-f]{2})*"
+hexstring = text .regexp "([0-9a-f]{2})+"
 ~~~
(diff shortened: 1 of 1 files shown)


rmarx:
There is of course a subtle difference between "the field was absent/empty" or "we have omitted the field in this log due to privacy/size reasons"... somewhat related to https://github.com/quicwg/qlog/issues/72.

I think this might be bigger than just the hexstring though (e.g., see the FIN bit on a STREAM frame, which we currently also just treat as optional to indicate it was present or not)

LPardue:
I think we should support empty. For instance packets using 0-length connection IDs.

Help me understand the change though. Does the PR propose allowing empty or restricting it?

rmarx:
@LPardue the PR restricts it. Currently, we do allow empty. 

As I said above and as discussed on the call, I don't really have a preference. 

For the CID case specifically though, I don't think anyone would actually log empty strings for the CID in each event. You can get the info that a 0-length CID was used from the long-header packet information, where the CID lengths are also logged. The `empty string` CIDs here would only be needed if you somehow didn't log the connection setup.

rmarx:
Discussed on call. No clear preference, so keep as-is.

Issue #173, opened 2021-08-06
RawInfo doesn't play well with UDP datagram headers


rmarx:
Currently, we re-use the RawInfo struct to give a simple re-usable way to log raw packet/frame/datagram contents.

In the struct, we have two length-fields: the full `length` and `payload_length`, assuming the header length can be calculated from those two. 

However, we do not have a way to log e.g., only the payload without the header (we can only log the full value truncated, but not indicate it starts at the payload).

This isn't a problem for QUIC/H3 packets/frames, but as @jlaine pointed out, it is a problem for UDP datagrams. In most stacks, you don't get the pure UDP packet header and so can't log it (so you'd always start with the payload).

This can be solved in several different ways:
1. disallow usage of RawInfo for datagrams
2. make explicit that datagram-related events start at the payload
3. log a fixed-value/fixed-size dummy UDP datagram (e.g., 8 bytes of zeroes)
4. extend RawInfo to make it more flexible

I'm currently most in favor of 2, as I don't see many people logging full UDP datagrams instead of logging the QUIC packets.

rmarx:
Fixed by #274

PR #274, opened 2023-01-10
Make use of raw and RawInfo more consistent

### LINKED ISSUE #54: Allow more wire-image indicators
Currently, we abstract out a lot of the on-the-wire specifics.

A good example is in the STREAM frame: there, the length and offset fields are optional, depending on how the frame should be handled. Simply having the same fields optional in qlog doesn't convey quite the same semantics: did the implementation simply not log them or were they not present from the start? 

More importantly though: if no length is set, the frame extends to the end of the packet, so it does have a length, which you'd probably want to log (that's the way it's currently designed), but so you loose the info that the length field wasn't encoded on the wire. 

This is just one example of similar problems across the board. I rather like the simplicity of the current setup, but we should have ways to signal the explicit wire image as well. 

cc @huitema

### LINKED ISSUE #174: Editorial update to RFCs/draft-34 + rawInfo + general consistency
since qlog started early on in the quic/H3 design process, we have some leftovers from early drafts in the docs.

We've updated most things over time as they were changed, but some aspects (e.g., most of the quic recovery events mention draft-23/27, `stream_state_updated` mentions draft-23, `packet_sent` triggers refer draft-23/19, etc.)

We need to go through each event and match it to terminology used in the final versions of the documents. 
For most things, this would just require removing mentions of old drafts, but in some cases this might require further changes. 

Finally, many events in the quic/H3 docs incorrectly use the pre-draft-02 way of logging raw information (e.g., as bytes fields directly), instead of properly using the new `raw:RawInfo` type. This should be updated.

Subsumes #46.

### LINKED ISSUE #243: Properly use RawInfo everywhere
Currently, most events use `RawInfo`, but some still use a separate `raw: hexstring` + `raw_length: uint32` setup. 

1. Figure out why (probably just forgot to update all when making the change)
2. Change to `RawInfo` where appropriate, or add comment why `RawInfo` isn't used

rmarx:
Fixes #243, #54 and #173.
Partially fixes #174.

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -115,26 +115,6 @@ identifier, potentially suffixed by the vantagepoint type (For example,
 abcd1234_server.qlog would contain the server-side trace of the connection with
 GUID abcd1234).
 
-## Raw packet and frame information
-
-This document re-uses the definition of the RawInfo data class from {{QLOG-MAIN}}.
-
-Note:
-
-: As HTTP/3 does not use trailers in frames, each HTTP/3 frame header_length can
-be calculated as header_length = RawInfo:length - RawInfo:payload_length
-
-Note:
-
-: In some cases, the length fields are also explicitly reflected inside of frame
-headers. For example, all HTTP/3 frames include their explicit payload lengths in
-the frame header. In these cases, those fields are intentionally preserved in the
-event definitions. Even though this can lead to duplicate data when the full
-RawInfo is logged, it allows a more direct mapping of the HTTP/3 specifications to
-qlog, making it easier for users to interpret. In this case, both fields MUST have
-the same value.
-
-
 # HTTP/3 and QPACK Event Overview
 
 This document defines events in two categories, written as lowercase to follow
@@ -255,10 +235,8 @@ Importance: Base
 Emitted when a stream's type becomes known. This is typically when a stream is
 opened and the stream's type indicator is sent or received.
 
-Note: most of this information can also be inferred by looking at a stream's id,
-since id's are strictly partitioned at the QUIC level. Even so, this event has a
-"Base" importance because it helps a lot in debugging to have this information
-clearly spelled out.
+stream_type_value is the actual, numerical code (the parsed VLIE-encoded
+integer).
 
 Definition:
 
@@ -270,7 +248,7 @@ HTTPStreamTypeSet = {
     stream_type: HTTPStreamType
 
     ; only when stream_type === "unknown"
-    ? raw_stream_type: uint64
+    ? stream_type_value: uint64
 
     ; only when stream_type === "push"
     ? associated_push_id: uint64
@@ -289,10 +267,9 @@ HTTPStreamType =  "request" /
 ## frame_created {#http-framecreated}
 Importance: Core
 
-HTTP equivalent to the packet_sent event. This event is emitted when the HTTP/3
-framing actually happens. Note: this is not necessarily the same as when the
-HTTP/3 data is passed on to the QUIC layer. For that, see the "data_moved" event
-in {{QLOG-QUIC}}.
+This event is emitted when the HTTP/3 framing actually happens. Note: this is
+not necessarily the same as when the HTTP/3 data is passed on to the QUIC layer.
+For that, see the "data_moved" event in {{QLOG-QUIC}}.
 
 Definition:
 
@@ -315,11 +292,9 @@ data_moved event.
 ## frame_parsed {#http-frameparsed}
 Importance: Core
 
-HTTP equivalent to the packet_received event. This event is emitted when the
-HTTP/3 frame is parsed. Note: this is not necessarily the same as when the
-HTTP/3 data is actually received on the QUIC layer. For that, see the
-"data_moved" event in {{QLOG-QUIC}}.
-
+This event is emitted when the HTTP/3 frame is parsed. Note: this is not
+necessarily the same as when the HTTP/3 data is actually received on the QUIC
+layer. For that, see the "data_moved" event in {{QLOG-QUIC}}.
 
 Definition:
 
@@ -333,11 +308,10 @@ HTTPFrameParsed = {
 ~~~
 {: #http-frameparsed-def title="HTTPFrameParsed definition"}
 
-Note: in HTTP/3, DATA frames can have arbitrarily large lengths to reduce frame
-header overhead. As such, DATA frames can span many QUIC packets and can be
-processed in a streaming fashion. In this case, the frame_parsed event is emitted
-once for the frame header, and further streamed data is indicated using the
-data_moved event.
+HTTP/3 DATA frames can have arbitrarily large lengths to reduce frame header
+overhead. As such, DATA frames can span multiple QUIC packets. In this case, the
+frame_parsed event is emitted once for the frame header, and further streamed
+data is indicated using the data_moved event.
 
 ## push_resolved {#http-pushresolved}
 Importance: Extra
@@ -406,10 +380,11 @@ $HTTPFrame /= HTTPBaseFrames
 {: #httpbaseframe-def title="HTTPBaseFrames definition"}
 
 ### HTTPDataFrame
+
 ~~~ cddl
 HTTPDataFrame = {
     frame_type: "data"
-    ? raw: hexstring
+    ? raw: RawInfo
 }
 ~~~
 {: #httpdataframe-def title="HTTPDataFrame definition"}
@@ -531,13 +506,14 @@ HTTPReservedFrame = {
 
 ### HTTPUnknownFrame
 
+frame_type_value is the actual, numerical code (the parsed VLIE-encoded
+integer).
+
 ~~~ cddl
 HTTPUnknownFrame = {
     frame_type: "unknown"
-    raw_frame_type: uint64
-
-    ? raw_length: uint32
-    ? raw: hexstring
+    frame_type_value: uint64
+    ? raw: RawInfo
 }
 ~~~
 {: #httpunknownframe-def title="UnknownFrame definition"}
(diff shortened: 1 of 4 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1083,63 +1083,43 @@ This section specifies such common definitions.
...
-behavior or determining encoding/framing overheads. However, these fields are not
-always necessary and can take up considerable space if logged for each packet or
-frame. They can also have a considerable privacy and security impact. As such,
-they are grouped in a separate optional field called "raw" of type RawInfo (where
-applicable).
+While qlog is a high-level logging format, it also allows the inclusion of most
+raw wire image information, such as byte lengths and byte values. This is useful
+when for example investigating or tuning packetization behavior or determining
+encoding/framing overheads. However, these fields are not always necessary, can
+take up considerable space and can have a considerable privacy and security
```
LPardue, line 1090:
```suggestion
take up considerable space, and can have a considerable privacy and security
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1083,63 +1083,43 @@ This section specifies such common definitions.
...
+impact. Where applicable, these fields are grouped in a separate optional field
```
LPardue, line 1091: worth xref to security considerations?

LPardue, line 1091:
```suggestion
impact. Where applicable, these fields are grouped in a separate, optional, field
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1083,63 +1083,43 @@ This section specifies such common definitions.
...
-(e.g., QUIC's AEAD tag), event definitions documents SHOULD define other ways of
-logging the trailer_length to make the header_length calculation possible.
-
-: The exact definitions entities, headers, trailers and payloads depend on the
-protocol used. If this is non-trivial, event definitions documents SHOULD include
-a clear explanation of how entities are mapped into the RawInfo structure.
-
-Note:
+The RawInfo:data field can be truncated for privacy or security purposes, see
+{{truncated-values}}. In this case, the length and payload_lenght fields should
```
LPardue, line 1115:
```suggestion
{{truncated-values}}. In this case, the length and payload_length fields should
```

LPardue, line 1115: is this a SHOULD requirement?

Either way, it might help to explain why keeping the non-truncated length is advised.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1083,63 +1083,43 @@ This section specifies such common definitions.
...
-as implementations might not necessarily employ the most efficient VLIE scheme for
-all values. As such, to make exact size-analysis possible, implementers should use
-explicit lengths in RawInfo rather than reconstructing them from other qlog data.
-Similarly, tool developers should only utilize RawInfo (and related information)
-in such tools to prevent errors.
+This document does not specify explicit header_length or trailer_length fields.
+In protocols without trailers, header_length can be calculated by subtracting
+the payload_length from the length. In protocols with trailers (e.g., QUIC's
+AEAD tag), event definitions documents SHOULD define a way to make the
+header_length calculation possible.
```
LPardue, line 1122:
```suggestion
AEAD tag), event definitions documents SHOULD define a how to support
header_length calculation.
```

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1076,25 +1078,17 @@ TransportDataMoved = {
...
 (for example to indicate data from QUIC CRYPTO frames being passed to a TLS
-library ("security") or from HTTP/3 to QPACK ("qpack")).
-
-Note: this event type is part of the "transport" category, but really spans all
-the different layers. This means we have a few leaky abstractions here (for
-example, the stream_id or stream offset might not be available at some logging
-points, or the raw data might not be in a byte-array form). In these situations,
-implementers can decide to define new, in-context fields to aid in manual
-debugging.
+library or from HTTP/3 to QPACK).
```
LPardue, line 1091: Is this paragraph necessary? There are no events with a `direction` field. The purpose of from/too fields is already desdcribed in the earlier paragraphs. This text seems superflous to me.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1824,8 +1814,9 @@ PathResponseFrame = {
 
 ### ConnectionCloseFrame
 
-raw_error_code is the actual, numerical code. This is useful because some error
-types are spread out over a range of codes (e.g., QUIC's crypto_error).
+error_code_value is the actual, numerical code (the parsed VLIE-encoded
+integer). This is useful because some error types are spread out over a range of
+codes (e.g., QUIC's crypto_error).
```
LPardue, line 1819: I find this a bit ambigous to read. How about something like

```suggestion
The error_code_value is the numerical value without VLIE encoding.
This is useful because some error types are spread out over a range of
codes (e.g., QUIC's crypto_error).
```

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1856,13 +1847,15 @@ HandshakeDoneFrame = {
 
 ### UnknownFrame
 
+frame_type_value is the actual, numerical code (the parsed VLIE-encoded
+integer).
```
LPardue, line 1851: Again, reads ambiguous to me.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1927,6 +1920,12 @@ TBD
 
 ## Since draft-ietf-qlog-quic-events-02:
```
LPardue, line 1921:
```suggestion
## Since draft-ietf-qlog-quic-events-03:
```

LPardue:
Looks good to me modulo a few nits and editorials

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -289,10 +266,9 @@ HTTPStreamType =  "request" /
 ## frame_created {#http-framecreated}
 Importance: Core
 
-HTTP equivalent to the packet_sent event. This event is emitted when the HTTP/3
-framing actually happens. Note: this is not necessarily the same as when the
-HTTP/3 data is passed on to the QUIC layer. For that, see the "data_moved" event
-in {{QLOG-QUIC}}.
+This event is emitted when the HTTP/3 framing actually happens. Note: this is
+not necessarily the same as when the HTTP/3 data is passed on to the QUIC layer.
```
LPardue, line 270:
```suggestion
This event is emitted when the HTTP/3 framing actually happens. This does
not necessarily coincide with HTTP/3 data getting passed to the QUIC layer.
```

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -333,11 +307,10 @@ HTTPFrameParsed = {
...
 
-Note: in HTTP/3, DATA frames can have arbitrarily large lengths to reduce frame
-header overhead. As such, DATA frames can span many QUIC packets and can be
-processed in a streaming fashion. In this case, the frame_parsed event is emitted
-once for the frame header, and further streamed data is indicated using the
-data_moved event.
+HTTP/3 DATA frames can have arbitrarily large lengths to reduce frame header
+overhead. As such, DATA frames can span multiple QUIC packets. In this case, the
+frame_parsed event is emitted once for the frame header, and further streamed
+data is indicated using the data_moved event.
```
LPardue, line 313: This seems like it can be genericised and moved somewhere else: there is no tight coupling between H3 and QUIC frame boundaries meaning a H3 frames could span many packets, or many frames could be in a single packet.

Implementations can deploy streaming strategies to any H3 frame parsing. I agree that DATA frames are the most likely, but really any frame could be larger than a packet. 

Furthermore, due to varint H3 frame fields and the lack of preserved boundaries, some implementations might generate data_moved events before frame_parsed events. Cloudlfare quiche does this, for example. 

The text is better than the old text, so lets not block the PR. I'll open a separate issue.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -886,23 +886,21 @@ TransportDatagramsSent = {
...
-Note: QUIC itself does not have a concept of a "datagram_id". This field is a
-purely qlog-specific construct to allow tracking how multiple QUIC packets are
-coalesced inside of a single UDP datagram, which is an important optimization
-during the QUIC handshake. For this, implementations assign a (per-endpoint)
-unique ID to each datagram and keep track of which packets were coalesced into the
-same datagram. As packet coalescing typically only happens during the handshake
-(as it requires at least one long header packet), this can be done without much
-overhead.
+The "datagram_id" is a qlog-specific concept to allow tracking of QUIC packet
+coalescing inside UDP datagrams. Implementations can assign a per-endpoint
```
LPardue, line 898: I realise this text was already there but why is it per-endpoint?

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -886,23 +886,21 @@ TransportDatagramsSent = {
...
+unique ID to each datagram, and reflect this in other events to track QUIC
+packets through processing steps.
+
+Since QUIC implementations rarely control UDP logic directly, the raw data
+excludes UDP-level headers in all fields.
```
LPardue, line 903: Consider moving this above the previous paragraph to match the ordering of the fields

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -918,27 +916,28 @@ TransportDatagramsReceived = {
...
-Note: for more details on "datagram_ids", see {{transport-datagramssent}}.
+For more details on "datagram_ids", see {{transport-datagramssent}}.
 
 ## datagram_dropped {#transport-datagramdropped}
 Importance: Extra
 
 When a UDP-level datagram is dropped. This is typically done if it does not
-contain a valid QUIC packet (in that case, use packet_dropped instead).
+contain a valid QUIC packet. If it does, but the QUIC packet is dropped for
+other reasons, {{transport-packetdropped}} should be used instead.
```
LPardue, line 934: This would just render as a section number right? That's not very reader-friendly.
```suggestion
other reasons, packet_dropped ({{transport-packetdropped}}) should be used instead.
```

draft-ietf-quic-qlog-quic-events.md
rmarx: Because there is no (obvious?) in-band way to share datagram-ids between individually logging endpoints? To keep them consistent, you'd have to have a stable hash or something?

LPardue (approved):

PR #279, opened 2023-01-13
Clean up text around events not belonging to a single connection

### LINKED ISSUE #35: Add guidance to server developers
@marten-seemann asked how to best approach logging from a server's perspective, given that things like version negotiation and stateless retry are not inherently tied to a single connection. We should add some informative guidance on how to best approach this to the spec, depending on how much state you're willing to keep around 

Some options:

1. low state: keep a separate qlog file for the entire server. This logs vneg, retry, etc.. Then, when a connection is truly accepted, start a new .qlog for the individual connection, containing all events thereafter. The server.qlog can then also contain an event signalling the acceptance of a new connection for later cross-linking between the files.
2. low state: keep a single huge qlog file for everything, using the "group_id" field to allow later de-multiplexing into separate connections (I believe quant does this atm)
3. stateful: if you already track vneg/retry and link them up with the final connection, you can output them in the per-connection qlog file as well

Maybe also shortly talk about some of the trade-offs in each option. Also talk about how to approach server-level events like server_listening and packet_dropped in separate scenarios.

rmarx:
Fixes #35


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -136,31 +136,22 @@ the QUIC specifications to qlog, making it easier for users to interpret.
...
-
+A single qlog event trace is typically associated with a single QUIC connection.
+However, for several types of events (for example, a {{transport-packetdropped}}
+event with trigger value of "connection_unknown"), it can be impossible to tie
+them to a specific QUIC connection, especially on the server.
+
+Implementations are free to choose how to handle these events. Some options include:
+
+* Log them in a separate endpoint-wide trace (or use a special group_id value)
+  not associated with a single connection. This can be more difficult to implement.
```
LPardue, line 147: grammar police: more difficult to implement than what?

If there are challenges or difficulties we can say that, but the "more" qualifier makes the sentence a comparison and this could be misinterpreted to mean that we are comparing this option to the others. If we are actually trying to compare the options (I don't think we are but I could be wrong) then we probably want to restructure a bit.

rmarx: Discussed on call. Best to just remove these parts here, unless we have concrete data/proof or examples.

LPardue (approved):

Issue #198, opened 2022-03-19
Logging events and dealing with clock drift


LPardue:
We had a question about quiche's logging on https://github.com/cloudflare/quiche/issues/1197 this got me thinking some more about representations of time in qlog.

In quiche, we use a monotonic non-decreasing clock. When a trace is started, we store a monotonic start time internally and don't put that in the qlog. As events are created, we calculate a relative time since the start and record that.

I've now realised this behaviour seems to differ from what the qlog draft states in https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#name-timestamps

> Relative: Specify a full "reference_time" timestamp (typically this is done up-front in "common_fields", see [Section 3.4.7](https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#common-fields)) and include only relatively-encoded values based on this reference_time with each event. The "reference_time" value is typically the first absolute timestamp. This approach uses a medium amount of characters.

The problem is that clock drift can occur. The spec mentions this briefly but I wonder if the problem is larger than people give consideration to. For the relative and base time formats, you'd potentially end up smaller or negative values, which might break qlog tooling.

I don't have much insight into solving clock drift problems but could we consider adding a monotonic time format type in addition?

LPardue:
@rmarx asked me to explain this further or more clearly. So lets try a different angle.

Here is what qlog says

> The "time" field indicates the timestamp at which the event occurred. Its value is typically the Unix timestamp since the 1970 epoch (number of milliseconds since midnight UTC, January 1, 1970, ignoring leap seconds). However, qlog supports two more succinct timestamps formats to allow reducing file size. The employed format is indicated in the "time_format" field, which allows one of three values: "absolute", "delta" or "relative".

Each event has a timestamp, whose value is calculated according to the format. qlog hints that the UNIX epoch is "typical", which leaves the door open to other epoch. But qlog doesn't really support expressing what the epoch is explicitly. 

The `relative` encoding is described as:

> specify a full "reference_time" timestamp (typically this is done up-front in "common_fields", see [Section 3.4.7](https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#common-fields)) and include only relatively-encoded values based on this reference_time with each event. The "reference_time" value is typically the first absolute timestamp. This approach uses a medium amount of characters.

[...]

PR #290, opened 2023-01-23
Rework the Timestamps section, add monotonic

LPardue:
Most of what was in here is good but we can tighten it up a bit.

After tightening up, it makes it easier to introduce another TimeFormat
- monotonic - that doesn't use an explicit reference_time.

Address the monotic request made in #198 but doesn't resolve the issue completely it because the discussion on clock drift is not expanded. I can add that if we like this direction.


LPardue:
also note I removed any specific example of "privacy-preserving timestamps". The old example about using system up time seems the antithesis of privacy :smile:

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -592,68 +592,86 @@ JSON serialization:
...
+relative:
+: A relative timestamp based on some relative time base.  The log SHOULD specify
+  a "reference_time" timestamp, either by providing it explicitly in the
+  "reference_time" field in "common_fields" ({{common-fields}}) or logging an
+  absolute time in the first event. This approach uses a medium amount of
+  characters.
+
+monotonic:
+: A relative timestamp based on an opaque monotonically nondecreasing clock.
+  The "reference_time" field MUST NOT be applied to this format, since it based
```
rmarx, line 632:
```suggestion
  The "reference_time" field MUST NOT be applied to this format, since it's based
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -592,68 +592,86 @@ JSON serialization:
...
+  on an opaque clock that cannot be interpreted beyond the process that
```
rmarx, line 633: This whole sentence is hard to understand (I'm not even sure I really grasp it), and the `it's` is ambiguous: is it "this format" or "the reference_time"?

Overall, imo this section does not really explain how monotonic is different from relative, nor why it's useful/needed.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -592,68 +592,86 @@ JSON serialization:
...
+
+Events in each individual trace SHOULD be logged in strictly ascending timestamp
+order (though not necessarily absolute value, for the "delta" format). Tools CAN
+sort all events on the timestamp before processing them, though are not required
+to (as this could impose a significant processing overhead). This can be a problem
+especially for multi-threaded and/or streaming loggers, who could consider using a
+separate post-processor to order qlog events in time if a tool do not provide this
+feature.
+
+There is no default base clock for formats that rely on a sharable clock. Tools
```
rmarx, line 644:
```suggestion
There is no default base clock for formats that rely on a shareable clock. Tools
```

rmarx: This is also a bit vague... what exactly is a "shareable clock", terminology not used before this. Does it mean sharing the same clock across traces (e.g., one taken at the server, one at the client, later combined?) or does it mean deriving timing values from a single clock reading (i.e., relative and delta). I assume the latter, but the quiche PR that triggered this is mainly talking about the former?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -592,68 +592,86 @@ JSON serialization:
...
+SHOULD NOT assume the ability to derive the absolute Unix timestamp from qlog
+traces. Tools should not rely events to be be consistent across traces, even
```
rmarx, line 646:
```suggestion
traces. Tools should not rely on timestamps to be consistent across traces, even
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -592,68 +592,86 @@ JSON serialization:
...
 
-The relative approach will:
-- set the reference_time to 1500 in "common_fields"
-- use: 0, 5, 22, 88
+Example "reference" TimeFormat values, assuming a reference_time of 1500
+in "common_fields":
+0, 5, 22, 88
+
+Example "monotonic" TimeFormat values:
+0, 5, 22, 88
```
rmarx, line 670: This enforces my confusion about why `monotonic` is needed and how it's different from `relative`... 

The way I understand it is that monotonic does not necessarily adhere to milliseconds between events? 
So say the relative values would be 
`0, 5, 22, 88`

Then the monotonic ones for the same absolute event timestamps could be
`0, 4, 17, 94`
(due to clock drift, intentional fuzzing for privacy/timing attacks, etc.)

If that's true, the examples should reflect this better.

rmarx (changes requested):
I largely agree with the rewrites of the existing text.

I however don't really see the need for the `monotonic` format. The text doesn't explain why it's useful, nor can I really find good arguments for it in the quiche PRs besides "clock drift". 

IIUC, monotonic is still a kind of relative, but you can't assume the events arrived at exactly `reference_time` + timestamp milliseconds. That's fine, but then we should imo just say something general about all formats like:

`timestamp values aren't always perfectly accurate, due to clock drift, implementation limitations, privacy or security considerations. They are only guaranteed to be monotonically increasing.`

That imo covers the load for the 3 existing formats, and allows you to use `relative` with a `reference_time` of 0 for monotonic (which, IIUC, is what quiche has been doing?)

draft-ietf-quic-qlog-main-schema.md
LPardue: Yeah I'm not a fan of the word shareable here but struggled to find a suitable term. This is really just intended to be a rewording pf the old text that was focussed on unix-y time. If folks are using an abosulute timestamp, somewhere they should be indicating the actual format and rules for that, including where its epoch is anchored. Saying "you don't have to use UNIX time" like the old text is glossing over a lot of complexity.

Happy to rewrite this part if we can agree on what should be communicated.

draft-ietf-quic-qlog-main-schema.md
LPardue: Relative and monotonic are both formats that use a delta encoding but that's where the similarity ends.

Per the current definition of reference format, it has to be related to something that various systems can agree on. For example, sticking January 1 2023 in all my logs and logging the duration in seconds since that date would work for a reference time. Ideally, you understand what that date is, so that you can add the time deltas in order determine their absolute timestamp. It is important to agree on a calendaring or date/time system because things like leap seconds, daylight savings, or clock skew can mean your relative-time event has negative values. This makes sorting qlogs fun.

With monotonic clocks, there is no universal calendaring or date/time. A system picks _something_ and you can  then calculate durations against that. As long a single system generates events, the durations are guaranteed to never decrement. See for example https://linux.die.net/man/3/clock_gettime where

> CLOCK_MONOTONIC
Clock that cannot be set and represents monotonic time since some unspecified starting point.

So one core difference is that reference can be negative and montonic can never.  So the reference example might want to include that.

Another core difference is that monotonic typically has no reference_time that can be communicated meaningfully. Its literally not possible to obtain it on such systems.

we could make the monotonic values different for the sake of it, if it helps avoid confusion.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -592,68 +592,86 @@ JSON serialization:
...
+: A relative timestamp based on some relative time base.  The log SHOULD specify
+  a "reference_time" timestamp, either by providing it explicitly in the
+  "reference_time" field in "common_fields" ({{common-fields}}) or logging an
+  absolute time in the first event. This approach uses a medium amount of
+  characters.
+
+monotonic:
+: A relative timestamp based on an opaque monotonically nondecreasing clock.
+  The "reference_time" field MUST NOT be applied to this format, since it's based
+  on an opaque clock that cannot be interpreted beyond the process that
```
LPardue, line 633:
```suggestion
  The "reference_time" field MUST NOT be applied to this format, since monotonic
  time is measured against an opaque clock that cannot be interpreted beyond the system that
```

draft-ietf-quic-qlog-main-schema.md
LPardue: I tried to clarify the sentence to make it clear what it being measured.

Beyond that, I'm not sure how to more clearly describe that `relative` and `monotonic` are different. The monotonic description is succinct. So the relative format probably needs more wordsmithing for preciseness, rather than have people making assumptions.

LPardue:
As disscussed on the issue and at IETF 116, we'll probably go a different way than this PR. Keeping it open for my own reference but don't merge

LPardue:
Retiring this in favor of #433

PR #291, opened 2023-03-01
Update description of datagram_id usage

### LINKED ISSUE #281: Revisit use of datagram_id
Currently, we have a mechanism for tracking QUIC packet coalescing in a single UDP datagram by means of the `datagram_id` field. This was originally added as per #91. 

Main intended use: when a UDP payload is received containing multiple coalesced QUIC packets, those QUIC packets are given the same "datagram ID" (which is a receiver-side chosen number that's unique per received datagram, which is needed because UDP datagrams don't have a packet number or similar). As such, when 2 separate packets are logged in 2 separate `packet_received` events, but they have the same value for the `datagram_id` field, it's clear they were coalesced. Similarly, 2 `packet_sent` events with the same `datagram_id` indicate sender-side coalescing. 

As the original issue #91 indicates, this can be deduced from the `datagram_*` events as well, but those are often too chatty to leave on. The `datagram_ids` fields on the `datagram_*` events are for completeness / ability to track coalesced packets across the different events. 

I'm personally not the biggest fan of this solution (did you eventually implement this in quic-go, @marten-seemann?), but I don't immediately see a better approach.

CC @LPardue as he indicated the use of `datagram_id` wasn't clear to him, which means we either need to revisit or clarify its use.

rmarx:
Fixes #281

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -904,9 +904,11 @@ Since QUIC implementations rarely control UDP logic directly, the raw data
 excludes UDP-level headers in all fields.
 
 The "datagram_id" is a qlog-specific concept to allow tracking of QUIC packet
-coalescing inside UDP datagrams. Implementations can assign a per-endpoint
-unique ID to each datagram, and reflect this in other events to track QUIC
-packets through processing steps.
+coalescing inside UDP datagrams. Since QUIC generates many UDP datagrams, unique
+identifiers are required to be able to track them individually in qlog traces.
+However, neither UDP nor QUIC exchanges datagram identifiers on the wire.
+Selecting identifier is values is thus left to qlog implementations, that should
+consider how to generate unique values within the scope of their created traces.
 
 ## datagrams_received {#transport-datagramsreceived}
 Importance: Extra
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -904,9 +904,11 @@ Since QUIC implementations rarely control UDP logic directly, the raw data
 excludes UDP-level headers in all fields.
 
 The "datagram_id" is a qlog-specific concept to allow tracking of QUIC packet
-coalescing inside UDP datagrams. Implementations can assign a per-endpoint
-unique ID to each datagram, and reflect this in other events to track QUIC
-packets through processing steps.
+coalescing inside UDP datagrams. Since QUIC generates many UDP datagrams, unique
+identifiers are required to be able to track them individually in qlog traces.
+However, neither UDP nor QUIC exchanges datagram identifiers on the wire.
+Selecting identifier is values is thus left to qlog implementations, that should
```
LPardue, line 910:
```suggestion
Selecting identifier values is thus left to qlog implementations, which should
```

LPardue (approved):
LGTM modulo nit

LPardue:
rebase on main should fix the cddl tests

PR #298, opened 2023-05-14
CDDL: use consistent spacing

lnicco:
This is a re-do of #253

PR #299, opened 2023-05-16
Add more whitespaces after CDDL comment lines (continues #298)

lnicco:


LPardue (approved):
Its a trivial problem. Let's just merge this as is and fixup any stragglers if and when we find any.

marten-seemann (approved):

Issue #178, opened 2021-10-04
support for DATAGRAMS in data_moved events


LPardue:
Slightly related to #124

https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#section-3.3.15 defines the data_moved event. The stream_id field is optional. It isn't clear if the type of data I'm moving is stream data (and I don't have the stream ID) or if its datagram data (and I'll never have the stream ID).

An non-exhaustive list of options to address this are:

1) leave it as is, and let tools assume that omitting the stream ID field implies datagram. (That breaks down for implementations that can't populate the stream ID for whatever reason)
2) add a field to describe the type of data moved
3) Be more clear that `data_moved` is only for stream data.Then possibly:
  a) rename the event to `stream_data_moved`
  b) add a new `datagram_data_moved` event

LPardue:
If we accept that we'll punt detailing DATAGRAMS in the base qlog specs, then any of these options would still fundamentally work. 

It think its a crapshoot between option 2 and 3, and comes down to ergonomics. Is it easier for consumers of qlog to index off an event type, or a field in an event? Or is it totally neutral?

PR #310, opened 2023-07-05
Add QUIC and HTTP/3 DATAGRAM frame definitions

rmarx:
As discussed at IETF 115 and 116, we will add support for most RFC-status QUIC and HTTP/3 extensions to the main documents.

This PR adds our intended support for RFCs 9221 and 9297. Note that this does NOT include full support for those entire documents. Specifically, we only add the QUIC and H3 DATAGRAM frames, not the full "capsule protocol", which was deemed out of scope for the core qlog documents. 

These changes were originally part of a separate document at https://github.com/rmarx/draft-marx-quic-qlog-datagram, which is now deprecated.

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -60,8 +60,9 @@ level schema defined in {{QLOG-MAIN}}.
 # Introduction
 
 This document describes the values of the qlog name ("category" + "event") and
-"data" fields and their semantics for HTTP/3 {{RFC9114}} and QPACK
-{{!QPACK=RFC9204}}.
+"data" fields and their semantics for the HTTP/3 protocol {{!HTTP3=RFC9114}},
+QPACK {{!QPACK=RFC9204}}, and some of their extensions (see
+{{!H3-DATAGRAM=RFC9297}}).
 
 > Note to RFC editor: Please remove the follow paragraphs in this section before
 publication.
@@ -195,9 +196,15 @@ H3ParametersSet = {
 }
 
 H3Parameters = {
+    ; RFC9114
     ? max_field_section_size: uint64
+
+    ; RFC9204
     ? max_table_capacity: uint64
     ? blocked_streams_count: uint64
+
+    ; RFC9297 (SETTINGS_H3_DATAGRAM)
+    ? h3_datagram: uint16
 
     ; additional settings for grease and extensions
     * text => uint64
@@ -206,7 +213,7 @@ H3Parameters = {
 {: #h3-parametersset-def title="H3ParametersSet definition"}
 
 This event can contain any number of unspecified fields. This allows for
-representation of reserved settings (aka grease) or ad-hoc support for
+representation of reserved settings (aka GREASE) or ad-hoc support for
 extension settings that do not have a related qlog schema definition.
 
 ## parameters_restored {#h3-parametersrestored}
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-qlog-h3-events.md
```diff
@@ -365,7 +372,8 @@ H3BaseFrames = H3DataFrame /
                H3GoawayFrame /
                H3MaxPushIDFrame /
                H3ReservedFrame /
-               H3UnknownFrame
+               H3UnknownFrame /
+               H3DatagramFrame
```
LPardue, line 376: stylistically, I'd put the datagram one directly after max push ID

LPardue: actually see my other comment on the issue

LPardue:
The QUIC frame is good. Howvever, in HTTP/3, there is not an actual datagram frame being exchanged. I think we need separate events such as "H3DatagramCreated" and "H3DatagramParsed"

LPardue:
Lets try and resolve https://github.com/quicwg/qlog/issues/178 in this PR

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -508,6 +516,19 @@ H3UnknownFrame = {
...
 {: #h3unknownframe-def title="UnknownFrame definition"}
 
+## H3DatagramFrame Definition
+
+The HTTP/3 DATAGRAM frame is defined in {{Section 2 of !RFC9297}}.
+
+~~~ cddl
+H3DatagramFrame = {
+    frame_type: "datagram"
+    quarter_stream_id: uint64
```
lnicco, line 526: should this have its own type with max 2^60-1 ?
I guess we don't do that for other fields e.g. the ones encoded with a QUIC Integer

rmarx:
> The QUIC frame is good. Howvever, in HTTP/3, there is not an actual datagram frame being exchanged. I think we need separate events such as "H3DatagramCreated" and "H3DatagramParsed"

Discussed on call: replace with something like this:

```
H3DatagramParsed = {
    quarter_stream_id: uint64
    
    ? payload: $H3DatagramPayload
    ? raw: RawInfo
}

$H3DatagramPayload = any
```

draft-ietf-quic-qlog-h3-events.md
lnicco: actually thinking more about this, I think it's better to keep the full range. 
This way it would be easier to spot if an implementation is not using the id correctly

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -60,8 +60,9 @@ level schema defined in {{QLOG-MAIN}}.
 # Introduction
 
 This document describes the values of the qlog name ("category" + "event") and
-"data" fields and their semantics for HTTP/3 {{RFC9114}} and QPACK
-{{!QPACK=RFC9204}}.
+"data" fields and their semantics for the HTTP/3 protocol {{!HTTP3=RFC9114}},
+QPACK {{!QPACK=RFC9204}}, and some of their extensions (see
```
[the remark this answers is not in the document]
rmarx: As discussed: will indeed list them all once individual PRs are merged (prevent merge conflicts).

rmarx:
> Lets try and resolve #178 in this PR

As discussed in the call, we'll go for option 3 there:

> Be more clear that data_moved is only for stream data.Then possibly:
> a) rename the event to stream_data_moved
> b) add a new datagram_data_moved event

LPardue (approved):
LGTM

PR #315, opened 2023-07-05
Remove .well-known retrieval mechanism

### LINKED ISSUE #313: Is the .well-known address really necessary?
The main schema defines a means for clients to download generated qlogs by accessing them at a well-known address.  It hints that this is via HTTP but doesn't mandate that or provide more details about the expected HTTP interaction.

I'm curious how many people actually support this. The spec itself says that `qlog implementers SHOULD NOT enable this .well-known endpoint in typical production settings`, anyone that follows that advice will not. If we axpect this to be more of a debugging methodology, that might suggest that sharing the location (e.g. URI or something else) using out-of-band mechanisms would be just as managable. 

The proposal is also one sided. There is no capability defined for a client to submit its own logs for a server to later analyse. Not sure if that's a real problem but it is a constraint someone might question later in the process.

Since the .well-known access does not run in the "hot path" of event logging, it would seem fine to extract this and punt it to some follow on work.

LPardue:
Closes #313

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1864,7 +1864,7 @@ hesitant to include these fields for all but the most stringent use cases.
 
 # IANA Considerations
 
-TODO: primarily the .well-known URI
+There are no IANA considerations.
 
 --- back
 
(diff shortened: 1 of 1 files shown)


rmarx (approved):

LPardue:
Gonna sit on this a little while we wait for the WG to consider it.

marten-seemann (approved):

PR #321, opened 2023-07-05
Tidy up discussion about event names

LPardue:
This section had some hangover text from when an event contained
separate category and type fields. That's confusing.

Lets be clearer and more explicity about how we expect the name field to
be used. We should probably add additional language about what to do if
a parser encouners a name field *is not* valid. But that can be done as
followup.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -546,46 +546,30 @@ Tools SHOULD NOT assume the ability to derive the absolute Unix timestamp from
 qlog traces, nor allow on them to relatively order events across two or more
 separate traces (in this case, clock drift should also be taken into account).
 
-### Category and Event Type {#name-field}
+### Event Names {#name-field}
 
-Events differ mainly in the type of metadata associated with them. To help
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -546,46 +546,30 @@ Tools SHOULD NOT assume the ability to derive the absolute Unix timestamp from
...
+field is an identifier that parsers can use to decide how to interpret the event
+metadata contained in the "data" field (see {{data-field}}).
+
+Event names indicate a category and type. The "name" field MUST contain a
+non-empty character sequence representing a category, followed by a colon (':'),
+followed by a non-empty non-empty character sequence representing a type.
 
 Category allows a higher-level grouping of events per specific event type. For
 example for QUIC and HTTP/3, the different categories could be "quic", "http",
 "qpack", and "recovery". Within these categories, the event Type provides
```
rmarx, line 561:
```suggestion
"qpack", and "recovery". Within these categories, the event type provides
```

rmarx (approved):
LGTM minus two nits

Issue #283, opened 2023-01-17
Versioning for additional schema


LPardue:
We presently have 3 documents: main schema, quic schema, HTTP/3 and QPACK schema.

We presently have 1 version field `qlog_version` to express changes in any of these documents.

We  make efforts to keep these things in lock step, but the relationship is quite implicit and not explicitly stated. That's not brilliant and adds some friction.

One proposal I have is to articulate the schema(s) used in the log. I don't think we need to go as far as XML does (e.g. https://www.oreilly.com/library/view/xml-in-a/0596007647/re168.html) but we could just include the I-D or RFC name as a string. Then each additional schema document can express what version of qlog main schema it depends on. I'll write this up as a PR.

rmarx:
As discussed on the call yesterday, everyone generally likes this concept on how to tackle versioning. Additionally, it can help replace `protocol_type` together with #286. 

One thing I personally did not really love in current PR #284 is that we use a separate approach for the main schema version (indicated via `qlog_version`) and the protocol event defintion versions (indicated via the new `additional_schema` field). Another option would be to treat the main schema as just another one in the list.

Put differently, this is what the PR currently proposes: 
```
"qlog_version": "0.3",
"qlog_format": "JSON",
"additional_schema": [
  "draft-ietf-quic-qlog-quic-events-03",
  "draft-ietf-quic-qlog-quic-h3-events-03"
],
```
While a more consistent option would be:
```
"qlog_format": "JSON",
"qlog_schemas": [
  "draft-ietf-quic-qlog-main-schema-04",
  "draft-ietf-quic-qlog-quic-events-03",
  "draft-ietf-quic-qlog-quic-h3-events-03"
],
```

Downsides I could find:
- We're stuck with the `qlog_schemas` field and its definition forever. No way to e.g., refactor it to a list of protocols and their versions (e.g., instead of `"draft-ietf-quic-qlog-quic-events-03"` it could say `["HTTP/3", "v5.1"]`). This is true for the `qlog_version` field as well, but that seems like much less of an issue.
- For parsers that can handle multiple different versions, it's easier to do a quick check on `qlog_version`, rather than having to parse the full `qlog_schemas` field to find the main-schema entry and decide if they support it or not.

I think those downsides are sufficient to go with the original proposal, but in that case, I'd rename `additional_schema` to `protocol_schemas` or `event_schemas` (or the `qlog_version` field to `qlog_schema_version` ;))

LPardue:
[...]

PR #329, opened 2023-07-25
additional_schema using URI/URN

LPardue:
alternative to https://github.com/quicwg/qlog/pull/284/files, based on the suggestion by @JonathanLennox to do something similar to https://www.rfc-editor.org/rfc/rfc8285.html#section-5 and https://www.iana.org/assignments/rtp-parameters/rtp-parameters.xhtml#rtp-parameters-10

Closes https://github.com/quicwg/qlog/issues/283

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -92,10 +92,14 @@ fields.
 
 # Overview
 
-This document describes how the HTTP/3 and QPACK can be expressed in qlog using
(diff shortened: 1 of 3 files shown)


rmarx:
I assume this is subsumed by #415 @LPardue? Should we close this already or wait for the other one to be merged?

LPardue:
I forgot about this entirely :facepalm:, it doesn't make sense to keep open as I don't want to have to fix the bitrot.

PR #331, opened 2023-10-21
Reflow of main schema document first half

LPardue:
The organic growth of the document has made, in my opinion, the flow
of the document a bit janky.

This is especially so for JSON-SEQ. After reading the whole document
that frames things as qlog - there's a text buried down in the
serialization that requires the reader to discard a lot of internalized
context. It shouldn't be like that because QLogFile and QLogFileSeq are
separate things that share a tonne of qlog types.

This change focuses on sections 3-5. We draw our the primary themes that
underpin qlog by picking out the terms that we actually use. Instead of
schema -> traces -> Field name semantics in one section, we move to
QLogFile schema and related traces and Tace, QlogFileSeq and related TraceSeq,
VantagePoint, Events, Raw packet info, and Common Events. Finishing up with
guidelines for people to define their own events.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -221,7 +221,8 @@ omitted the default value of "JSON" applies.
 The optional "title" and "description" fields provide additional free-text
 information about the file.
 
-The optional "traces" field contains an array of qlog traces ({{trace}}), each of which contain metadata and an array of qlog events ({{events}}).
+The optional "traces" field contains an array of qlog traces ({{trace}}), each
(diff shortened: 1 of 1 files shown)


LPardue:
This is pretty hard to diff in git. See https://author-tools.ietf.org/api/iddiff?url_1=https://quicwg.github.io/qlog/draft-ietf-quic-qlog-main-schema.txt&url_2=https://quicwg.github.io/qlog/main-spec-reflow/draft-ietf-quic-qlog-main-schema.txt for soemthing easier

rmarx:
Merging this with a few fixed nits. Sorry for not following proper review procedure, but it's draft cutoff day ;)

PR #336, opened 2023-11-08
Add path_id and Connection Migration events

rmarx:
This PR adds basic support for associating paths (defined by IP + port + Connection ID) with individual qlog events. This is done through a textual per-event `path_id` which can be associated with the necessary metadata using the `path_assigned` event (to prevent the `path_id` itself having to contain encoded information/becoming too long).

While using the new path association possibilities in combination with logged PATH_CHALLENGE/PATH_RESPONSE frames (and other info like preferred_address) is enough to implicitly track connection migration attempts, it can be useful to explicitly log migration progression as well. For this, the `migration_state_updated` event was also added.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -710,7 +710,7 @@ could be serialized as
 
 ## Path {#path-field}
 
-A qlog event can be associated with a single network "path" (usually, but not
+A qlog event can be associated with a single "network path" (usually, but not
 always, identified by a 4-tuple of IP addresses and ports). In many cases, the
 path will be the same for all events in a given trace, and does not need to be
 logged explicitly with each event. In this case, the "path" field can be omitted
@@ -718,7 +718,7 @@ logged explicitly with each event. In this case, the "path" field can be omitted
 "common_fields" instead (see {{common-fields}}).
 
 However, in some situations, such as during QUIC's Connection Migration or when
-using MultiPath features, it is useful to be able to split events across
+using Multipath features, it is useful to be able to split events across
 multiple (concurrent) paths.
 
 Definition:
@@ -730,7 +730,7 @@ PathID = text .default ""
 
 
 The "path" field is an identifier that is associated with a single network path.
-This document intentionally does not further define how to choose this
+This document intentionally does not define further how to choose this
 identifier's value per-path or how to potentially log other parameters that can
 be associated with such a path. This is left for other documents. Implementers
 are free to encode path information directly into the PathID or to log
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -496,6 +496,7 @@ Event = {
     time: float64
     name: text
     data: $ProtocolEventBody
+    ? path: PathID
     ? time_format: TimeFormat
     ? protocol_type: ProtocolType
     ? group_id: GroupID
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -707,6 +708,36 @@ could be serialized as
 ~~~~~~~~
 {: #data-ex title="Example of the 'data' field for a QUIC packet_sent event"}
 
+## Path {#path-field}
+
+A qlog event can be associated with a single network "path" (usually, but not
```
qdeconinck, line 713: I would either suggest quoting "network path" or remove the network word.
```suggestion
A qlog event can be associated with a single "network path" (usually, but not
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -707,6 +708,36 @@ could be serialized as
...
+always, identified by a 4-tuple of IP addresses and ports). In many cases, the
+path will be the same for all events in a given trace, and does not need to be
+logged explicitly with each event. In this case, the "path" field can be omitted
+(in which case the default value of "" is assumed) or reflected in
+"common_fields" instead (see {{common-fields}}).
+
+However, in some situations, such as during QUIC's Connection Migration or when
+using MultiPath features, it is useful to be able to split events across
```
qdeconinck, line 721: Nit :-)
```suggestion
using Multipath features, it is useful to be able to split events across
```

qdeconinck (approved):
I think such PathId introduction, as well as the related events, seem like a good way to log multipath events, regardless on how the final design will evolve (we could be more specific about the content of PathID later if required).

LPardue:
We have a few issues related to migration, which ones does this PR address? Asking because it would be good to close things we think will be fixed.

LPardue (approved):
:shipit:

rmarx:
So, I'm looking to merge this by Monday (draft cutoff for IETF 119).

I see that @qdeconinck and @mpiraux have been doing some implementation work on this in quiche; has that led to additional insights (especially on how "ready for multipath" this design is)? Any major changes needed? 

Additionally, I could use some help to make sense of @mirjak's comments at the last meeting (not because they're not good comments, I just don't understand enough of the intricacies of paths/multipath probably).

Here is the specific part: https://youtu.be/qbAm_HfLv_c?feature=shared&t=4841 where she talks about needing not just the `path_id` when receiving an ACK because you also need to have a CID to identify a multipath packet number. I follow that, BUT I thought that was just always taken care of by the `ACK_MP` frame (which includes the CID? https://quicwg.org/multipath/draft-ietf-quic-multipath.html#name-ack_mp-frame) so this I'm confused... 

I also feel you could have path_ids like "pathx:CID1" and "pathx:CID2" (so kind of hierarchical path_ids that link to the same conceptual "path" but differentiate based on the CID being used). If that kind of scheme is canonified by a later "qlog for multipath" it would ensure consistency across implementations (if the issue is that the hierarcharical approach is too loosy-goosy for now ;)). 

In short: I'm planning to merge this by Monday unless we have major blockers here (that of course doesn't mean this can't change later, but I don't want to keep this as a floating PR too much longer). Thanks!

mirjak:
The current multipath draft doesn't have an explicit path ID. You say in the text a path is identified by the 4-tuple. This is not strictly aways true as due to migration the 4 tuple can change while the path is effectively the same. In the current draft we omit this problem by not defining an explicit path ID but _only_ use the CID as identifier.

However having this said we are still discussing that problem and there is a PR that does introduce an explicit path ID. If that's the thing that ends up being standardised in the end, your text is mostly fine. 

If that is not the solution we go for in the multipath draft, I recommend to not introduce a potentially ambiguous path ID and only use the CID as identifier. 

Sorry for not being more helpful at this point.

rmarx:
Thanks for the info @mirjak 

So, I understand there's no direct path ID in multipath, which is why I also de-couple the `pathID` in qlog to be any string, and implementations/specs are free to fill that in as they like. If there is a "proper" `pathID` introduced in multipath later, that can just be used. 

If not, the real question here is if implementations can (somewhat) easily derive a textual pathID to uniquely identify a path by coupling it to the info in `path_assigned`. As you say, a path is not always just the 4-tuple, so we also have an array of ConnectionIDs in `PathEndpointInfo`, which IIUC SHOULD be enough to cover that. Similarly, this allows keeping the same `pathID` even if the 4 tuple changes after migration (just use a new `path_assigned` event with the same `pathID` but changing the other parameters). 

> If that is not the solution we go for in the multipath draft, I recommend to not introduce a potentially ambiguous path ID and only use the CID as identifier.

There, I would say that you can use assign the CID to `pathID` and use `path_assigned` to give additional information on the path. Would that seem reasonable? 


The real question here is if we have enough to be usable, no matter what the multipath draft ends up doing (as having actual/full multipath support with everything properly worked out is a non-goal at this point).

PR #343, opened 2023-11-17
add a packet_dropped trigger for duplicate packets

marten-seemann:
See section 12.3 of RFC 9000:

> A receiver MUST discard a newly unprotected packet unless it is certain that it has not processed another packet with the same packet number from the same packet number space. Duplicate suppression MUST happen after removing packet protection for the reasons described in Section 9.5 of [QUIC-TLS].


LPardue (commented):
LGTM

PR #344, opened 2023-11-17
Update references to old drafts to point to RFCs

LPardue:
Relates to https://github.com/quicwg/qlog/issues/174.

None of these seem controversal


marten-seemann (approved):
LGTM, but I haven't checked that all the chapter numbers are correct.

PR #353, opened 2023-11-30
Restructure and reflow security section

### LINKED ISSUE #259: Security and Privacy: general structure and content
The version of the [Security and Privacy Considerations in main-schema-draft-04](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-main-schema-04.html#name-security-and-privacy-consid) was intended to be a first proposal that can be iterated upon.

In particular, @martinthomson has already [proposed an alternative approach](https://github.com/quicwg/qlog/pull/252#pullrequestreview-1152379427):

> This is a lot of text. I like the enumeration of the types of risk, but I fear that the extrapolation from there is both incomplete and a little long-winded.
>
> The focus on anonymization is not one that I would emphasize. I know that [Christian has toyed with the concept](https://huitema.wordpress.com/2020/07/21/scrubbing-quic-logs-for-privacy/), but that was idle speculation, not anything rigorous.
> 
> In the end, this text should be very clear that qlog files present a privacy risk to those participating in the connection. They should therefore be treated accordingly. Any notion that these files might be sanitized in a way that might remove that risk should be changed to focus more on defense in depth. That is, look at access control and retention limits as primary means of protection, not anonymization.
> 
> I would instead structure this roughly as follows:
> 
> - A brief introductory piece.
> - An articulation of the types of privacy risk that might be presented by capture, storage, and use of qlog data. This part is fairly lengthy and could probably be a subsection.
> - A description of the operational implications and potential actions that might be taken first from a security standpoint, then for privacy remediation. qlog files that are records of interactions with peers are data that those peers (or users) potentially have rights over (I like this point, so make it prominently, but do not refer to laws; do not refer to the GDPR especially).
> 1. Access controls required to enable qlog or to expand the information that is captured.
> 2. Access controls required to access stored qlog files.
> 3. Limits to retention of qlog files.
> 4. Finally, observe that limiting what is captured or anonymization might be possible, but that this could either be insufficient from a privacy perspective, reduce the usefulness of the data, or (highly likely) both.
> 
> It's easy to focus on the point about not being able to link qlog data with people. But it's safer to treat that as an open research problem than invite implementations and deployments to indulge in amateur research on the topic.

This issue serves to collect further thoughts and comments from the WG post IETF-115. It is a continuation of #142. 

_Note that we are explicitly trying to keep this part of the qlog documents quite concise. A deeper investigation on how to properly handle sensitive data in logs is a more general topic that we hope to tackle in the IETF in the near future. These initial qlog docs should provide just enough guidance to make users aware of the risk and to recommend best practices._

LPardue:
This is an attempt to implement some of Martin Thomson's proposed changes.

It promotes some of the discussion into a general brief intro. Then
speaks concretely about the data the qlog can put at risk. Then it
presents access control considerations, followed by data minimization.

Per Martin's suggestion, data minimization for addressing risks is
played down a little. I ended up changing this text the most but tried
to shift, incorporate, or dedupe as much as I could and avoid deletion.

Closes #259

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1576,7 +1576,8 @@ be carried in qlog data:
 ## Operational implications and recommendations
 
 Considerations for operational implications should focus on log capture and
-access.
+access. Logging of Internet protocols is equivalent to access to store or read
+plaintext communications, so security considerations apply similarly to qlog.
 
 It is recommended that the ability to enable qlog capture is subject to access
 control and auditing. Furthermore, access control policy should extend to the
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1587,13 +1587,13 @@ more-sensitive data requires higher privileges).
 It is recommended that access to stored qlogs is subject to access control and
 auditing.
 
-Access control techniques in end user client environments can be limited.
-An end-user that might enable logging without understanding the implications of
-that choice on their privacy and security. Implementations should
-consider how to make enabling qlog conspicous, and resistant to automation or
-drive-by attacks. Examples include, requiring explicit actions to start a
-capture and isolation or sandboxing of capture from other activities in the
-same process or component.
+Access control techniques in end user client environments can be limited. An
+end-user that might enable logging without understanding the implications of
+that choice on their privacy and security. Implementations should consider how
+to make enabling qlog conspicous, and resistant to social engineering,
+automation, or drive-by attacks. Examples include, requiring explicit actions to
+start a capture, and isolation or sandboxing of capture from other activities in
+the same process or component.
 
 It is recommended that data retention policies are defined for the storage of
 qlog files.
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1517,7 +1517,7 @@ still provide adequate output for incomplete logs.
 Protocols such as TLS {{?RFC8446}} and QUIC {{?RFC9000}} provide varying degrees
 of secure protection for the wire image {{?RFC8546}}. There is inevitably
 tension between security and observability, when logging can reveal aspects of
-the wire image, that would ordinarily be protected. This tension equally applies
+the wire image that would ordinarily be protected. This tension equally applies
 to any privacy considerations that build on security properties, especially if
 data can be correlated across data sources.
 
@@ -1586,11 +1586,11 @@ It is recommended that access to stored qlogs is subject to access control and
 auditing.
 
 Access control techniques in end user client environments can be limited.
-End-users have the possibility of capturing their own logs and risking their own
-security and privacy in ways that could be surprising. Implementations should
+An end-user that might enable logging without understanding the implications of
+that choice on their privacy and security. Implementations should
 consider how to make enabling qlog conspicous, and resistant to automation or
 drive-by attacks. Examples include, requiring explicit actions to start a
-capture, and isolation or sandboxing of capture from other activities in the
+capture and isolation or sandboxing of capture from other activities in the
 same process or component.
 
 It is recommended that data retention policies are defined for the storage of
@@ -1602,11 +1602,12 @@ It is recommended that qlog files are encrypted in transit and at rest.
 
 Data minimization or anonymization techniques can be applied to qlog. While this
 can go some way to addressing security and privacy risks, implementers and
-operators are advised that the techniques might not provide a sufficient level
-of mitigation. Equally, minimization or anonymization can reduce the
-effectiveness of qlog data.
+operators are advised that removing or anonymizing data without sufficient care might
+not improve privacy or securrity.
+Minimization or anonymization might greatly reduce the usefulness
+of qlog data.
 
-Operator and implementors need to balance logging fields against the potential
+Operator and implementers need to balance the value of logged data against the potential
 risks inherent in their (involuntary) disclosure. This balance depends on the
 use case at hand (e.g., research datasets might have different requirements to
 live operational troubleshooting).
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1514,12 +1514,22 @@ still provide adequate output for incomplete logs.
 
 # Security and privacy considerations {#privacy}
 
-Protocols such as TLS {{!RFC8446}} and QUIC {{!RFC9000}} provide varying degrees
+Protocols such as TLS {{?RFC8446}} and QUIC {{?RFC9000}} provide varying degrees
 of secure protection for the wire image {{?RFC8546}}. There is inevitably
 tension between security and observability, when logging can reveal aspects of
 the wire image, that would ordinarily be protected. This tension equally applies
 to any privacy considerations that build on security properties, especially if
 data can be correlated across data sources.
+
+Any data that is determined to be necessary for a use case at hand could be
+logged or captured. As per {{?RFC6973}}, operators must be aware that such data
+will be at risk of compromise and this, in turn, risks the privacy of all
+entities that participate in logged internet protocol exchanges. Entities that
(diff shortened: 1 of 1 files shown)


LPardue:
The text diff might make this easier to review - https://author-tools.ietf.org/api/iddiff?url_1=https://quicwg.github.io/qlog/draft-ietf-quic-qlog-main-schema.txt&url_2=https://quicwg.github.io/qlog/sec-n-priv-mt/draft-ietf-quic-qlog-main-schema.txt

@martinthomson if you could find some time to review this it would be appreciated but we won't block on that.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1714,13 +1714,23 @@ still provide adequate output for incomplete logs.
 
 # Security and privacy considerations {#privacy}
 
-Protocols such as TLS {{!RFC8446}} and QUIC {{!RFC9000}} provide varying degrees
+Protocols such as TLS {{?RFC8446}} and QUIC {{?RFC9000}} provide varying degrees
 of secure protection for the wire image {{?RFC8546}}. There is inevitably
 tension between security and observability, when logging can reveal aspects of
 the wire image, that would ordinarily be protected. This tension equally applies
```
martinthomson, line 1720:
```suggestion
the wire image that would ordinarily be protected. This tension equally applies
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1714,13 +1714,23 @@ still provide adequate output for incomplete logs.
...
 to any privacy considerations that build on security properties, especially if
 data can be correlated across data sources.
 
+Any data that is determined to be necessary for a use case at hand could be
+logged or captured. As per {{?RFC6973}}, operators must be aware that such data
```
martinthomson, line 1725: This first sentence is not precise about whether it refers to the logging use case or the use case of the logged exchange.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
-
-To reduce the risk of data abuse after compromise, data can be anonymized,
-pseudonymized, otherwise permutated/replaced, truncated, (re-)encrypted, or
-aggregated. A partial discussion of applicable techniques (especially for IP
-address information) can be found in {{Appendix B of !DNS-PRIVACY=RFC8932}}.
-Operators should, however, be aware that many of these techniques have been
-shown to be insufficient to safeguard user privacy and/or to protect user
-identity, especially if a qlog data set is large or easily correlated against
-other data sources.
+Operational implications qlog should focus on log capture and access.
```
martinthomson, line 1778: This doesn't appear to be a complete and well-formed sentence.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+
+It is recommended that the ability to enable qlog capture is subject to access
+control and auditing. Furthermore, access control policy should extend to the
+level of information that is allowed to be captured (e.g., capture of
+more-sensitive data requires higher privileges).
```
martinthomson, line 1783: Would it be reasonable to suggest that any authorization for logging access be considered equivalent to plaintext access to communications.  I realize that some people might feel uncomfortable about that, because they enable logging where they would not enable plaintext access, but I think that considering them equivalent -- absent strong justification to the contrary -- better reflects the true nature of most communications.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+
+It is recommended that access to stored qlogs is subject to access control and
+auditing.
+
+Access control techniques in end user client environments can be limited.
+End-users have the possibility of capturing their own logs and risking their own
+security and privacy in ways that could be surprising. Implementations should
```
martinthomson, line 1790: This statement is a bit obtuse.  I would instead say:

```suggestion
An end-user that might enable logging without understanding the implications of
that choice on their privacy and security. Implementations should
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+consider how to make enabling qlog conspicous, and resistant to automation or
+drive-by attacks. Examples include, requiring explicit actions to start a
```
martinthomson, line 1792: ... or social engineering?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+capture, and isolation or sandboxing of capture from other activities in the
```
martinthomson, line 1793:
```suggestion
capture and isolation or sandboxing of capture from other activities in the
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+It is recommended that data retention policies are defined for the storage of
+qlog files.
+
+It is recommended that qlog files are encrypted in transit and at rest.
+
+## Data minimization or anonymization
+
+Data minimization or anonymization techniques can be applied to qlog. While this
+can go some way to addressing security and privacy risks, implementers and
+operators are advised that the techniques might not provide a sufficient level
```
martinthomson, line 1805:
```suggestion
operators are advised that removing or anonymizing data without sufficient care might
not improve privacy or securrity.
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+of mitigation. Equally, minimization or anonymization can reduce the
+effectiveness of qlog data.
```
martinthomson, line 1807:
```suggestion
Minimization or anonymization might greatly reduce the usefulness
of qlog data.
```

It is rare that the effect will be equal.  (You probably meant "at the same time" or "in the same way")

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+
+Operator and implementors need to balance logging fields against the potential
```
martinthomson, line 1809:
```suggestion
Operator and implementers need to balance the value of logged data against the potential
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1763,54 +1773,71 @@ be carried in qlog data:
...
+risks inherent in their (involuntary) disclosure. This balance depends on the
+use case at hand (e.g., research datasets might have different requirements to
+live operational troubleshooting).
```
martinthomson, line 1812: It seems like the difference in these examples is WHO has access to the data and the level to which the different entities might be trusted, so maybe this isn't so much about use cases.

martinthomson (commented):
Very happy to see that you don't recommend k-anonymity as a measure here.  Though maybe that is what you imply with the call out to anonymization.  There are some useful results from differential privacy in that regard, but that comes with some odd consequences for usability.

martinthomson:
Oh, I forgot to add one more point.  I've said this before, but I will again.

This is very long.  I would encourage you to find ways to have this section include fewer words, not more.

draft-ietf-quic-qlog-main-schema.md
LPardue: Yes, I'll update the text to reflect that.

draft-ietf-quic-qlog-main-schema.md
LPardue: yep that was my implication but I'll add that

draft-ietf-quic-qlog-main-schema.md
LPardue: I agree. I'm going to track this in a separate ticket (along with your other mention of use-cases)

LPardue:
In the interest of momentum, I'll punting the cutting down of text to https://github.com/quicwg/qlog/issues/358

PR #356, opened 2023-12-09
fix(quic-events): typo version_negotiation on *un*-supported version

mxinden:
The server sends a version_negotiation packet when it does *not* support the client version.

Also remove duplicate "of".

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -494,10 +494,10 @@ The `version_information` event supports QUIC version negotiation; see {{Section
 6 of QUIC-TRANSPORT}}. It has Core importance level; see {{Section 9.2 of
 QLOG-MAIN}}.
 
-QUIC endpoints each have their own list of of QUIC versions they support. The
+QUIC endpoints each have their own list of QUIC versions they support. The
 client uses the most likely version in their first initial. If the server does
-support that version, it replies with a version_negotiation packet, containing
-supported versions. From this, the client selects a version. The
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -494,10 +494,10 @@ The `version_information` event supports QUIC version negotiation; see {{Section
 6 of QUIC-TRANSPORT}}. It has Core importance level; see {{Section 9.2 of
 QLOG-MAIN}}.
 
-QUIC endpoints each have their own list of of QUIC versions they support. The
+QUIC endpoints each have their own list of QUIC versions they support. The
 client uses the most likely version in their first initial. If the server does
-support that version, it replies with a version_negotiation packet, containing
-supported versions. From this, the client selects a version. The
+not support that version, it replies with a version_negotiation packet,
+containing supported versions. From this, the client selects a version. The
```
marten-seemann, line 500: Not introduced in this PR, but taking the opportunity to fix another nit: The packet is called Version Negotiation packet, according to RFC 8999.

```suggestion
not support that version, it replies with a Version Negotiation packet, which
contains its supported versions. From this, the client selects a version. The
```

marten-seemann (commented):
Thank you!

draft-ietf-quic-qlog-quic-events.md
mxinden: Thanks. Today I learned that RFC 9000 is not the first QUIC related RFC.

marten-seemann (approved):

PR #359, opened 2023-12-09
Define uint64 duality once

### LINKED ISSUE #229: Section on I-JSON seems to confusingly redefine uint64
The section on I-JSON notes serialization concerns about 64-bit numbers that might be required for protocols like QUIC or HTTP/3.

It then says:

> Concretely, the following definition of uint64 should override the original and (web-based) tools should take into account that a uint64 field can be either a number or string.

> uint64 = text / uint .size 8

I find this confusing. This document defines a custom `unit64` itself. It's not entirely clear when the above override applies or not.

LPardue:
Closes #229

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -139,11 +139,15 @@ clarity:
 ~~~ cddl
 ; CDDL's uint is defined as being 64-bit in size
 ; but for many protocol fields it is better to be restrictive
-; and explicit
+; and explicit. Some environments cannot support representation
+; of the full range of allowed 64-bit values, so text is also
+; defined; see Section 10.1.1.
+
 uint8 = uint .size 1
 uint16 = uint .size 2
 uint32 = uint .size 4
-uint64 = uint .size 8
+uint64 = uint .size 8 /
+         text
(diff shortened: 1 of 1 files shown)


marten-seemann (commented):
I don’t like this change. The CDDL is supposed to be an independent representation of the data, and now the quirks of one serialization are bleeding into this.

LPardue:
That's fair. But that's effectively what we've already been saying all along - it was just buried in a section about I-JSON. We have an interop issue if people miss this section. E.g. with no way to signal I-JSON, if I pass you a JSON qlog file with a large GREASE value encoded with a string and you try to parse it as a uint64 ... 💥. A quiet fail is also bad, being unable to represent data values (or to parse them) is a way to hide data within a connection that might be expected to be observable.

Things are leaky. CDDL explains that in https://www.rfc-editor.org/rfc/rfc8610#appendix-E. Maybe there's some other way to explain that in the docs and I'm open to it but I don't have much inspiration.

rmarx:
Discussed during editors meeting. 
Lucas will update this with new opinion about how we should handle uint64 in combo with CDDL.

LPardue:
overtaken by #388

PR #360, opened 2023-12-09
Reframe JSON interop concerns seperately of I-JSON

### LINKED ISSUE #339: I-JSON things
#289 does an editorial refactor, which hightlights that I-JSON doesn't really belong as a subsection of JSON serialization. In his review, Robin states:

> Since we removed truncation from this, it's a bit weird to have just a single subsection under JSON. It's also weird we don't explicitly say anything about "qlog_format"/file extension/MIME type when employing I-JSON (it's still just "JSON", but that should be explicitly spelled out imo).

> It makes more sense to promote I-JSON to a `##` here, so we have 3 sections for the 3 formats (this is also how they're introduced atm).

_Originally posted by @rmarx in https://github.com/quicwg/qlog/pull/289#discussion_r1120305282_

LPardue:
Since we added JSON-SEQ, it's been hard to square the circle about
where I-JSON fits in the document structure. I think that's because
actually the discussion should be framed in the context of interop
issues related to JSON messages. Our 64-bit values pose a problem
and the way CDDL suggest to solve that doesn't work - so we lay it
all out here.

There is obvious overlap here with PR #359. Whichever lands first
(if they even do) will cause some conflict that is easy to fixup.

Closes #339

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1226,10 +1226,10 @@ with general-purpose qlog tools.
 Considering these tradeoffs, JSON-based serialization formats provide features
 that make them a good starting point for qlog flexibility and interoperability.
 For these reasons, JSON is a recommended default and expanded considerations are
-given to how to map qlog to JSON ({{format-json}}, its subset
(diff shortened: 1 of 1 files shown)


LPardue:
This change looks big but its mostly just promoting the `I-JSON` section from 10.1.1 to a broader `JSON Interoperabiltiy` section 10.3

rmarx (approved):

PR #363, opened 2023-12-09
Move commentary on HTTP/3 frame and QUIC packetization

### LINKED ISSUE #280: Commentary on H3 frames and data_moved events
The DATA frame section includes some commentary on H3, frames and streams. In reviewing a PR, I made the following observations:

> This seems like it can be genericised and moved somewhere else: there is no tight coupling between H3 and QUIC frame boundaries meaning a H3 frames could span many packets, or many frames could be in a single packet.

> Implementations can deploy streaming strategies to any H3 frame parsing. I agree that DATA frames are the most likely, but really any frame could be larger than a packet. 

> Furthermore, due to varint H3 frame fields and the lack of preserved boundaries, some implementations might generate data_moved events before frame_parsed events. Cloudflare quiche does this, for example. 

_Originally posted by @LPardue in https://github.com/quicwg/qlog/pull/274#discussion_r1069621693_

LPardue:
Closes #280

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -166,6 +166,16 @@ layer, and there isn't always a one to one mapping between HTTP and QUIC events.
 The exchange of data between the HTTP and QUIC layer is logged via the
 "stream_data_moved" and "datagram_data_moved" events in {{QLOG-QUIC}}.
 
+HTTP/3 frames are transmitted on QUIC streams, which allows them to span
+multiple QUIC packets. Some implementations might send a single large frame,
+rather than a sequence of smaller frames, in order to amortize frame header
+overhead. HTTP/3 frame headers are represented by the frame_created
+({{h3-framecreated}}) and frame_parsed ({{h3-frameparsed}}) events. Subsequent
+frame payload data transfer is indicated by stream_data_moved events.
+Furthermore, stream_data_moved events can appear before frame_parsed events
+because implementations need to read data from a stream in order to parse the
+frame header.
+
 ## parameters_set {#h3-parametersset}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-h3-events.md
```diff
@@ -166,6 +166,16 @@ layer, and there isn't always a one to one mapping between HTTP and QUIC events.
 The exchange of data between the HTTP and QUIC layer is logged via the
 "stream_data_moved" and "datagram_data_moved" events in {{QLOG-QUIC}}.
 
+HTTP/3 frames are transmitted on QUIC streams, which allows them to span
+multiple QUIC packets. Some implementations might send a single large frame,
+rather than a sequence of smaller frames, in order to amortize frame header
```
marten-seemann, line 171: Do we need to specify that this only applies to DATA frames?

LPardue: Ah I see the disjoint in my previous understanding of why the text was the way it was. In theory extension frames could do something similar but nothing concrete is in scope, only DATA benefits from that usage.

Frames can span packets and there might be some value is explaining that here. Let me try to tweak the text a bit more.

rmarx: I actually think the text might be fine as-is? There's also the case where you have the H3 frame headers in Packet N and the actual payload in Packet N+1 (say you have multiple streams being packed into the same packet for some reason), so not tying this down to `DATA` only seems fine.

If you really want, you could change it to `Additionally, some implementations might send a single large frame,
rather than a sequence of smaller frames...`, so it's a bit clearer that the "single large frame" is a common occurrence of this issue but not the sole reason :)

rmarx (approved):

PR #383, opened 2024-01-28
Convert H3 setting param types to a real type

### LINKED ISSUE #382: H3Settings is a bit underspecified
https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html#section-5.3.4

```
H3Setting = {
    name: text
    value: uint64
}
```

Looking at IANA https://www.iana.org/assignments/http3-parameters/http3-parameters.xhtml, I think qlog nudges implementations to log the setting name. When an implementation receives an unknown setting (not a greased one) it won't know the name. It could log the setting type value code in the name field, but that is not specified. This makes things annoying for parsers since they need to do some thinking. 

One suggestion is to add another field like `type` that could contain the numerical type value. It would be duplicative to have both, so that might require making `type` and `name` optional.

LPardue:
Closes #382

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -558,11 +558,14 @@ H3CancelPushFrame = {
 
 ### H3SettingsFrame
 
-The `name_bytes` field supports logging the raw value of a setting identifier,
-to support logging unknown settings.
-Where an individual setting is not known, the name value of "unknown"
-can be used and the value captured in the raw_type field; a
-numerical value without variable-length integer encoding.
+The settings field can contain zero or more entries. Each setting has a name
+field, which corresponds to Setting Name as defined (or as would be defined if
+registered) in the the "HTTP/3 Settings" registry maintained at
+<https://www.iana.org/assignments/http3-parameters>.
+
+An endpoint that receives unknown settings is not able to log a specific name.
+Instead, the name value of "unknown" can be used and the value captured in the
+`name_bytes` field; a numerical value without variable-length integer encoding.
 
 ~~~ cddl
 H3SettingsFrame = {
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -574,12 +574,10 @@ H3SettingsFrame = {
 }
 
 H3Setting = {
-    ? name: text
-    ? name_bytes: uint64
     ? name: H3SettingsName
-
     ; only when name === "unknown"
-    ? raw_type: uint64
+    ? name_bytes: uint64
+
     value: uint64
 }
 
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -560,6 +560,9 @@ H3CancelPushFrame = {
 
 The `name_bytes` field supports logging the raw value of a setting identifier,
 to support logging unknown settings.
+Where an individual setting is not known, the name value of "unknown"
+can be used and the value captured in the type_value field; a
+numerical value without variable-length integer encoding.
 
 ~~~ cddl
 H3SettingsFrame = {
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-h3-events.md
```diff
@@ -521,16 +521,31 @@ H3CancelPushFrame = {
...
     frame_type: "settings"
     settings: [* H3Setting]
 }
 
 H3Setting = {
-    name: text
+    ? name: H3SettingsName
+
+    ; only when name === "unknown"
+    ? type_value: uint64
```
rmarx, line 538: Nit: Not sure I like `type_` here... it isn't called that in 9114 (there it's more `identifier`) and to align with `name` above and other similar fields, I'd go with `name_value` maybe?

LPardue: `name` or `identifier` seems fine to me. I was checking the IANA table which uses slightly different terms and it threw me off

LPardue:
Instead of a separate raw_type field, we'll just allow the nae to be a union of enum and uint64.

rmarx:
Heads up that I changed some of this in #417, by adding the `name_bytes` field (to be consistent with other places where we allow logging raw values), so that would replace the `raw_type` field proposed here.

LPardue:
> Heads up that I changed some of this in #417, by adding the `name_bytes` field (to be consistent with other places where we allow logging raw values), so that would replace the `raw_type` field proposed here.

So I've rebased and resolved the conflict. However, ISTM on a previous editor call we wanted to make this is union instead, so something like

```
H3Setting = {
    ? name: H3SettingsName / uint64

    value: uint64
}
```

and obviously a text update to describe how to pick which one. 

Do we still want to do that?

rmarx:
So, we've gradually moved away from that pattern IIRC (e.g., packet_number is now always a uint64 instead of uint64 / string to deal with IJSON). 

Looking through the other events, there are a few events that use this pattern (ResetStreamFrame, StopSendingFrame, ConnectionClosed, IPAddress), e.g., connectionclosed:

```
    ? connection_code: $TransportError /
                       CryptoError /
                       uint32
    ? application_code: $ApplicationError /
                        uint32
```

But most events that need something like this, use the 

```
field: string
field_bytes: hexstring
```

OR

```
field: string
field_bytes: uint64
```

It seems like ideally we'd just have a single way of dealing with these. I'm leaning more towards the explicit extra `field_bytes` pattern, to make things explicit that this is (likely) an "unexpected" value that's likely to cause problems + to prevent tools from having to special case types for fields (at least most of the time). 

If you agree, then we should merge this PR as-is (or with my proposed changes in the upcoming review ;)) and then I can make a new PR adding new `XYZ_bytes` fields to the non-conformant events.

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -568,10 +574,20 @@ H3SettingsFrame = {
 }
 
 H3Setting = {
-    ? name: text
+    ? name: H3SettingsName
```
rmarx, line 577:
```suggestion
    ? name: $H3SettingsName
```

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -568,10 +574,20 @@ H3SettingsFrame = {
...
+    ; only when name === "unknown"
     ? name_bytes: uint64
+
     value: uint64
 }
+
+H3SettingsName = "settings_qpack_max_table_capacity" /
```
rmarx, line 584:
```suggestion
$H3SettingsName /= "settings_qpack_max_table_capacity" /
```

rmarx (changes requested):
Make sure that the setting names can also be extended by future docs (before this was just `text` so was implicit, but now with the enum, we need to be explicit with a CDDL type socket)

LPardue:
I'm more than happy to go with this direction right now. Suggestions merged so I think GTG

rmarx (approved):
Looks good indeed. Feel free to merge, I'll open another issue to change the others and then later a PR to actually do so as well :)

PR #384, opened 2024-01-28
Tweak trigger_frame_type field and add it to connection_closed

### LINKED ISSUE #366: Improve connectivity:connection_closed

connectivity:connection_closed:

  This is missing the frame type field in the QUIC CONNECTION_CLOSE
  frame which is rather useful information.

  connection_code can be a string or a uint32, but QUIC uses a variable
  length integer here and thus this can be up to 62 bits. Same for the
  application_code.

  It is up to the decoder to infer by presence of connection_code or
  application_code to infer whether this was an application or transport
  CONNECTION_CLOSE. But this requires it to log such a value. It would
  be nice if this can be communicated without necessarily knowing the
  actual value. Consider either adding a field to distinguish between
  transport and app errors or replacing the "error" field of "trigger"
  with "app_error" and "transport_error".

LPardue:
Tweaks the language related to the ConnectionCloseFrame trigger_frame_type
and adds it to connection_closed event.

Closes #366

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -311,6 +311,12 @@ these however, there can be internal errors that occur that may or may not get
 mapped to the official error codes in implementation-specific ways. As such,
 multiple error codes can be set on the same event to reflect this.
 
+When the connection is closed due a connection-level error, the
+`trigger_frame_type` field can be used to log the frame that triggered the
+error. For known frame types, the appropriate string value is used. For unknown
+frame types, the numerical value without variable-length integer encoding is
+used.
+
 ~~~ cddl
 ConnectivityConnectionClosed = {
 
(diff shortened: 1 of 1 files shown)


rmarx:
While I like the general prose editorial work, I'm personally not too convinced that we need the `trigger_frame_type` on the `connectivity` event. As stated elsewhere, you rarely get all combined information from a single qlog event, and you often need to look at surrounding events for more low-level info (in this case: a field specifically on the CONNECTION_CLOSE frame). 

While I'm not terribly opposed to it, I'd still like to not add to the `connectivity` event, but do keep the editorial work with the CONNECTION_CLOSE frame (and maybe just reference that for the `connectivity` event?)

LPardue:
> While I'm not terribly opposed to it, I'd still like to not add to the `connectivity` event, but do keep the editorial work with the CONNECTION_CLOSE frame (and maybe just reference that for the `connectivity` event?)

Makes sense, I spun it off to its own PR https://github.com/quicwg/qlog/pull/386

rmarx:
As explained above, I think we should close this without action (i.e., don't add `trigger_frame_type` to `connection_closed`)

Issue #375, opened 2024-01-19
Improve quic:stream_data_moved


rmarx:
As reported by @hlandau on the mailing list:

quic:stream_data_moved:

  A brief explanation of each from/to value here and the intended meaning would
  be a big improvement. Also clarify that raw here refers to the referenced
  stream application data (and does not include frame headers etc.).

  Consider having a way to report when the end-of-stream (FIN) condition is
  reported to an application. One option here is to have an optional "fin" field
  here.

  There does not appear to be any way to communicate when a stream reset event is
  delivered to an application in an equivalent way, which would be useful.

rmarx:
While I can do the first part here (explain the from/to intents), I think @LPardue is more qualified to help with how we best indicate other signals have been delivered from transport to application.

I know that has a long history of discussion in general, and I personally don't think this should be part of the `data_moved` event (i.e., would need a new event in my opinion)

LPardue:
There's already a lot of text in these sections, personally I don't think explaining each field is going to be much more help. 

Re: `raw` raking it clear it is stream data not frames is a couple of words and worth it. 

I can see some value in the ability to indicate when an app requests a stop_sending or a reset or when wither is passed up to an app. I think that's a new event though.

For stream fin, I could argue a few ways. I think a `? fin: bool` field on stream_data_moved is probably the best tradeoff.

rmarx:
Discussed in editors meeting.
Robin: not the biggest fan of adding `fin` + some additional fields for stop_sending and reset_stream, but option.
Marten and Lucas: similar APIs in implementation: fin is possibly separate from the others (which are passed as errors). Errors often don't get communicated directly when happen, but only when application tries to read. So it does make sense to have it in `stream_data_moved` maybe. 
Lucas: will propose a PR to add these signals to `stream_data_moved`

PR #395, opened 2024-02-19
State clearly what lengths mean for stream and datagram data moved

LPardue:
Addresses part of #375

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1172,8 +1172,6 @@ length values in the `length` or `raw` fields MUST reflect that.
 
 For example, it can be useful to understand when data moves from an
 application protocol (e.g., HTTP) to QUIC stream buffers and vice versa.
-Similarly, when data moves from the application protocol layer into a
-user-facing application such as a web browser.
 
 The `stream_data_moved` event can provide insight into whether received data on
 a QUIC stream is moved to the application protocol immediately (for example per
@@ -1211,8 +1209,7 @@ QUICStreamDataMoved = {
     * $$quic-streamdatamoved-extension
 }
 
-$DataLocation /=  "user" /
-                  "application" /
+$DataLocation /=  "application" /
                   "transport" /
                   "network"
 
@@ -1233,8 +1230,7 @@ This event relates to datagram data only. There are no packet or frame headers a
 length values in the `length` or `raw` fields MUST reflect that.
 
 For example, passing from the application protocol (e.g., WebTransport) to QUIC
-Datagram Frame buffers and vice versa. Similarly, when data moves from the
-application protocol layer into a user-facing application such as a web browser.
+Datagram Frame buffers and vice versa.
 
 The `datagram_data_moved` event can provide insight into whether received data
 in a QUIC Datagram Frame is moved to the application protocol immediately (for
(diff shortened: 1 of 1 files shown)


rmarx:
As discussed during editorial meeting:
I'm not the biggest fan of current text since it isn't clear for all the possible uses (e.g., `transport` to `application` strips QUIC level stuff but keeps H3 frame headers, while `application` to `user` also strips those out etc.). Maybe best to word it in terms of "only the payload is moved to the higher layer" or something like that?

rmarx (changes requested):
See my latest comment in the thread (which I forgot to submit as an actual review, apparently)

LPardue:
> I'm not the biggest fan of current text since it isn't clear for all the possible uses (e.g., transport to application strips QUIC level stuff but keeps H3 frame headers, while application to user also strips those out etc.).

This is the QUIC document, so the intent of the proposed text

> There are no packet or frame headers and length values in the `length` or `raw` fields MUST reflect that.

is that it is solely in the scope of QUIC packets.

Now you mention it, I don't actually think the `user` location is suitable for the `$DataLocation` type. QUIC streams and datagrams don't mean anything without an application protocol. And how an application protocol uses them is specific to its definition. To use HTTP/3 as an example, I'm never going to pass QUIC stream or datagram data directly to a user. I'm going to pass them:

* HTTP request or response messages
* HTTP datagrams
* GOAWAY

There are HTTP/3 qlog events for each of these. 

Even an extension like WebTransport is going to pass WebTransport stream data, not raw QUIC. In which case, qlog for that extension should probably definine its own event type too.

rmarx:
> This is the QUIC document, so the intent of the proposed text is that it is solely in the scope of QUIC packets.

That's conceptually true, but doesn't really help us here? it's still unclear what this means when you pass data from `application` to `user`, so it doesn't solve the original issue in #375 .

--- 

Then, IIUC, your solution to that specifically is to remove the `user` option in the `from` and `to` and only keep it to `network, transport, application` since there it's clear what should/shouldn't be included. 

You argue that `application` to `user` doesn't make much sense for `quic:stream_data_moved` in the first place, since you're not passing QUIC stuff up to user, but application stuff. That is of course true.

The problem there is that we don't have something like `h3:data_moved` at this point in the HTTP/3 doc. We only have `h3:frame_parsed`, which doesn't necessarily indicate when the data is upstreamed to `user`. 

The original design of `transport:data_moved` was to be re-usable across all of this (so it really wasn't intended to be `quic:stream_data_moved` but instead more like `generic:data_moved`). 

---

With all that said, I can see 4 options:

1. Keep the current *_data_moved events in `quic` but make clear they can be used for app-level data as well but that they only pass the PAYLOAD of lower layers, not all packet/frame/whatever headers (my original proposal above)
2. Remove `user` as an option here and create an equivalent `h3:data_moved` event that has 3 targets: `user`, `application`, and `transport`. Not sure if we need an `h3:datagram_data_moved` and `h3:stream_data_moved` split here...
3. Remove `user` as an option here and just leave it at that (so have no explicit way of indicating when data was passed from H3 to the user application). If needed, that can be added in a later extension. 
4. Move the *_data_moved events to the generic category in the main document so they're re-usable across protocols. This is more akin to the original design intent, but gets annoying because you'd remove the `stream_id` field for example and lose the distinction between `stream_data_moved` and `datagram_data_moved`.

I currently lean towards either 1 or 3.

LPardue:
I went for option 3

> Remove user as an option here and just leave it at that (so have no explicit way of indicating when data was passed from H3 to the user application). If needed, that can be added in a later extension.

and pushed a commit. I think this is the simplest way to resolve the matter,

PR #396, opened 2024-02-19
Add fin and reset_stream fields

### LINKED ISSUE #375: Improve quic:stream_data_moved

LPardue:
Fixes #375.

Per the discussion, it can be useful to know when an application triggers a
stream close, or when it receives one, without having to do offset/length maths.

I omitted stop_sending in this commit because it seems a little bit of a stretch,
mainly because it's wire image has no offset or final size. Also, STOP_SENDING
relates to asking the peer to do something, which further complicates the
(potential) use of offset and length fields.

An alternative would be to keep `fin` here but define new events for reset and
stop. Those could both contain from, to and error fields but only the reset
would have offset and length.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1157,12 +1157,12 @@ processed first and afterwards the application layer reads from the streams with
 newly available data). This can help identify bottlenecks, flow control issues,
 or scheduling problems.
 
-The `fin` and `reset_stream` fields support optional logging of information
+The `additional_info` field supports optional logging of information
 related to the stream state. For example, an application layer that moves data
-into transport and simultaneously ends the stream, can set `fin` to true. As
+into transport and simultaneously ends the stream, can log `fin_set`. As
 another example, a transport layer that has received an instruction to reset a
-stream can indicate this to the application layer using the `reset_stream`
-field. In both cases, the length-carrying fields (`length` or `raw`) can be
+stream can indicate this to the application layer using `reset_stream`.
+In both cases, the length-carrying fields (`length` or `raw`) can be
 omitted or contain zero values.
 
 This event is only for data in QUIC streams. For data in QUIC Datagram Frames,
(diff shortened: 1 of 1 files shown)


rmarx:
Discussed during editorial meeting:
Keep `fin` here, move `reset_stream` and `stop_sending` to separate event(s) (possibly having 2 because reset_stream is so different from stop_sending).

@marten-seemann it would help to have a quick mockup of what `reset_stream_at` might look like (not because we want to add it to main qlog yet, but if the current design would be compatible with that, it would be nice :)

rmarx:
Actually discussed further during meeting, opinions changed :) 

I'm now in favor of something quite general, say:

```
? additional_info: [+ text]

; possible values for QUIC/H3 currently are "fin_set", "stream_reset", "stream_reset_at", "stop_sending"
```

With the idea that other events can be used to determine what exactly happened (e.g., look at the actual STREAM_RESET frame) and keep this event what it was originally intended for: showing when which type of data (and now info) is communicated between the layers (as this doesn't always coincide with packet sent/received)

rmarx (changes requested):
See latest comment (which I, again, didn't submit as an actual review apparently)

LPardue:
> See latest comment (which I, again, didn't submit as an actual review apparently)

Do you mean this one https://github.com/quicwg/qlog/pull/396#issuecomment-1968725064, or something else?

rmarx:
that one indeed :) sorry for the confusion

LPardue:
Reading between the lines, I think we don't want a `text` type but extendable socket-mabobs. Lemme give that a crack.

rmarx (approved):

PR #400, opened 2024-02-29
Make Parameters events extensible

### LINKED ISSUE #176: Define means for strong extensibility of TPs
https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-parameters_set defines some of the core QUIC TPs and then says

> Additionally, this event can contain any number of unspecified fields. This is to reflect setting of for example unknown (greased) transport parameters or employed (proprietary) extensions.

That's fine but is it weak and doesn't help interoperability much. We know this is going to be an area of innovation - new extensions and grease values. The spec would be improved by having a well-defined "extension bucket" field that TPs can go in, an array of integers tuples would be fine, for example.

This would help something like the qlog crate accomodate different users https://github.com/cloudflare/quiche/blob/master/tools/qlog/src/lib.rs#L1712 and parse input from different loggers.

rmarx:
Fixes #176.

This uses CDDL's "group socket" (https://datatracker.ietf.org/doc/html/rfc8610#section-3.9) to allow others to later specify additional parameters that can be logged to the event in a typed/explicit way. 

Example in a later document (say for reliable stream resets):

```
$$quic-parametersset-extension //= (
  reliable_stream_reset: text
)
```

One conceptual "downside" is that by definition all fields in the extension will be optional, even if not preceded by a `?` (as in the example above). This is because the extension isn't yet known when processing the current definition, so it has to be either `?` (optional) or `*` (0 or more, which is used here as it's recommended by RFC8610). I don't see this as a major issue, but something to be aware of.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ endif
 cddl:
 	@for f in $(drafts_source); do \
 	    echo "Validating $$f"; \
-	    ./cddl_validate.sh $$f > /tmp/foo 2>&1 ; \
+	    ./cddl_validate.sh $$f > /tmp/foo-$$f 2>&1 ; \
 	    if [ $$? -eq 0 ]; then \
 	        echo "  OK"; \
 	    else \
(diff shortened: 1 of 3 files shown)


LPardue (approved):
LGTM but I didn't really check the scripts that closely

PR #401, opened 2024-02-29
Enforce stricter extensibility for ProtocolEventData

### LINKED ISSUE #192: Rework $ProtocolEventBody
The new CDDL approach uses a CDDL extension point to be more strict about which values are allowed in the `Event:data` field. 

Currently, it's defined as:

```
; The ProtocolEventBody is any key-value map (e.g., JSON object)
; only the optional trigger field is defined in this document
$ProtocolEventBody /= {
    ? trigger: text
    * text => any
}
; event documents are intended to extend this socket by using:
; NewProtocolSpecificEvents = EventType1 / EventType2 / ... / EventTypeN
; $ProtocolEventBody /= NewProtocolSpecificEvents
```

However, this effectively still allows just about any JSON object as ProtocolEventBody, as we're just extending the $ProtocolEventBody.

If we want to be (slightly) more strict, we could do something like this:
```
; main doc
ProtocolEventBodyPrototype = {
    ? trigger: text
    * text => any
}

ProtocolEventBody = ProtocolEventBodyPrototype / $ProtocolEventBodies

; event docs
NewProtocolSpecificEvents = EventType1 / EventType2 / ... / EventTypeN
$ProtocolEventBodies /= NewProtocolSpecificEvents
```

Alternatively, we can just NOT allow the general form (the `ProtocolEventBodyPrototype`) and just include that as an example and require event bodies to provide at least 1 defined event.

rmarx:
Fixes #192.

Currently, `$ProtocolEventData` accepts just about anything, which makes it extensible but very useless for any real validation.

By not explicitly assigning the `* text => any` to the type in the main doc, we keep the extensibility, but improve validation capabilities.

This also properly adds the Generic and Simulation events to the `$ProtocolEventData`, which they really should have been from the start ;)

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -663,19 +663,20 @@ enable automated validation of aggregated qlog schemas.
 The only common field defined for the data field is the `trigger` field,
 which is discussed in {{trigger-field}}.
 
-~~~ cddl
+~~~
 ; The ProtocolEventData is any key-value map (e.g., JSON object)
-; only the optional trigger field is defined in this document
-$ProtocolEventData /= {
+; For example:
+EventTypeX = {
     ? trigger: text
-    * text => any
+   * text => any
 }
-; event documents are intended to extend this socket by using:
-; NewProtocolEventData = EventType1 /
-;                        EventType2 /
-;                        ... /
-;                        EventTypeN
-; $ProtocolEventData /= NewProtocolEventData
+
+; Event documents are intended to extend this socket by using:
+NewProtocolEventData = EventType1 /
+                       EventType2 /
+                       ... /
+                       EventTypeN
+$ProtocolEventData /= NewProtocolEventData
 ~~~
 {: #protocoleventdata-def title="ProtocolEventData definition"}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -663,19 +663,20 @@ enable automated validation of aggregated qlog schemas.
 The only common field defined for the data field is the `trigger` field,
 which is discussed in {{trigger-field}}.
 
-~~~ cddl
+~~~
 ; The ProtocolEventData is any key-value map (e.g., JSON object)
-; only the optional trigger field is defined in this document
```
LPardue, line 668: I recall we had some previous discussion about this. I'd never acknowledged this trigger field (from a serializer or parser perspective). Are we aware of anyone using it?

LPardue: To clarify, up till now every instance of a `$ProtocolEventData`, such as parsing a H3 frame https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html#section-4.6, had an optional `trigger` field. This is not something I ever implemented for quiche. For serialization, that's no problem because the field can be omitted in logs. But for parsing, it is a problem, because I would not deal well with a qlog file that included such triggers.

That's my faul for omitting it from the quiche implementation in the first place. I support this change removing `trigger`, it means my implementation will be correct. However, anyone that is currently making use of `trigger` in event bodies may be surprised by this change. Hence the ask, is anyone using it?

If they are, it would be good to know if its inside a few events, or many. We can always add an explicit trigger to events where it makes sense. I don't think it makes sense for there to be a blanket trigger.

rmarx: Discussed during editors meeting: 
@LPardue will send an email to the mailing list detailing this change and asking for feedback before merging.

rmarx: Email was sent May 13th (https://mailarchive.ietf.org/arch/msg/quic/e4CnmK2_NUrNAHOlo9Q-fV1JW6g/) without comments from the WG. So it seems this should be a go.

LPardue (approved):

rmarx:
Was subsumed by #417 (in a very similar fashion), closing without action.

PR #402, opened 2024-03-01
Clarify serialization expectations across documents

### LINKED ISSUE #357: Standardization formats intentions vs recommendations
Not sure I'm the biggest fan of referring explicitly to the other 2 docs here, since this MAIN doc is supposed to be generic (except maybe for some examples).

_Originally posted by @rmarx in https://github.com/quicwg/qlog/pull/289#discussion_r1420675552_

See my followup comments on that PR. It is useful to provide explicit statements but perhaps we want to extract the mention of QUIC and H3 event schema serialization intentions into their own documents. We can recommend in the base schema that each extension state if it is generic or make some format-specific optimization.

rmarx:
Fixes #357.

I'm not particularly fond of the prose in the main schema, but I think it covers the load for what I'd prefer to say (instead of linking to the current subdocuments directly). 

CC @LPardue


LPardue (approved):
LGTM

PR #403, opened 2024-03-01
Add clarification on is_coalesced and add it to more events

### LINKED ISSUE #370: Add clarification for is_coalesced to packet_sent and packet_received

quic:packet_sent:

  The meaning of `is_coalesced` should be clarified. I assume this is
  set to true unless a packet is both the first and last packet in a
  datagram. However that is not necessarily obvious when a packet is
  being generated; another option is to define this as being true for
  every packet but the first packet in a datagram.

quic:packet_received:

  See comment on is_coalesced above

rmarx:
Fixes #370. 

This better explains the intent behind `is_coalesced` and also adds it to other events that *probably* should have it as well. 

I'm not sure anymore that this is the best design however, since there's a lot of duplication for a feature that isn't THAT important to track...

I wonder if we can get away with something like "If implementations don't want to track individual datagram IDs, but still want to indicate a packet was/could be coalesced, they can use datagram_id = -1 to indicate this" (and then of course we remove `is_coalesced` and make `datagram_id` an `int32` instead of `uint32`).

That would be a bit dirty... but reduce the confusion and duplication considerably?

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -716,8 +716,15 @@ The `encryption_level` and `packet_number_space` are not logged explicitly:
 the `header.packet_type` specifies this by inference (assuming correct
 implementation)
 
-For more details on `datagram_id`, see {{quic-datagramssent}}. It is only needed
-when keeping track of packet coalescing.
+Multiple QUIC packets can be coalesced in the same UDP datagram, especially
+during the handshake (see {{Section 12.2 of QUIC-TRANSPORT}}). For this purpose,
+implementations SHOULD use qlog Datagram IDs and assign coalesced QUIC packets
+the same Datagram ID in the `datagram_id` field, see also
+{{quic-datagramssent}}. When multiple `packet_sent` events have the same
+`datagram_id`, this indicates they were coalesced into a single UDP datagram.
+However, in cases where implementations cannot track datagrams in this way, the
+`is_coalesced` field can be used to indicate that a packet was (intended to be)
+coalesced. Implementations MAY use both methods concurrently.
 
 ## packet_received {#quic-packetreceived}
 
@@ -751,8 +758,8 @@ The `encryption_level` and `packet_number_space` are not logged explicitly: the
 `header.packet_type` specifies this by inference (assuming correct
 implementation)
 
-For more details on `datagram_id`, see {{quic-datagramssent}}. It is only needed
-when keeping track of packet coalescing.
+For more details on `datagram_id` and `is_coalesced`, see {{quic-packetsent}}.
+They is only needed when keeping track of packet coalescing.
 
 ## packet_dropped {#quic-packetdropped}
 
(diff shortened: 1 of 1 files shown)


LPardue:
I think the new text that clarifies how the datagram_id is used is good. I don't think we benefit from the `is_coalesced` field, since it's a subset of a subset i.e. if an implementation cares enough about logging coalescing, then it can do the work to add datagram_id support.

rmarx:
Thanks for the comments on #370 @HLandau and @LPardue.

I agree and have removed `is_coalesced` in my latests commit. I've also moved all the text related to this to the `datagramssent` event so it's less spread-out. 

PTAL and see if the explanation there around `datagram_id` is clear enough. Thanks :)

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -881,14 +881,17 @@ QUICDatagramsSent = {
...
+excludes UDP-level headers in all RawInfo fields.
+
+Multiple QUIC packets can be coalesced in a single UDP datagram, especially
+during the handshake (see {{Section 12.2 of QUIC-TRANSPORT}}). However, neither
+QUIC nor UDP themselves provide an explicit mechanism to track this behaviour.
+Implementations willing to track coalescing across packet-level and
+datagram-level qlog events SHOULD thus use the local and qlog-specific concept
+of a "datagram identifier" in the `datagram_id` field. Selecting specific and
+locally unique `datagram_id` values is left to qlog implementations, but in
+general multiple packet-level events sharing the same `datagram_id` SHOULD
```
LPardue, line 893: This second SHOULD kudt seems to be repeating the requirement above?

rmarx: It is; I wasn't aware that's a problem. Would you prefer to change the first `SHOULD` to a `should` or the second one here?

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -881,14 +881,17 @@ QUICDatagramsSent = {
...
+indicate they were coalesced in the same UDP datagram.
```
LPardue, line 894: So I might be misreading what you intend but with current text but I could even go further with something like

```suggestion
qlog defines it's own mechanism for tracking coalescing across packet-level and
datagram-level qlog events, a "datagram identifier" carried in `datagram_id` fields.
qlog implementations that want to track coalescing can use this mechanism, the 
selectiion of specific and locally-unique `datagram_id` values is an implementation
choice.
```

draft-ietf-quic-qlog-quic-events.md
LPardue: I made a suggestion PTAL

hlandau:
Looks good. Might want to rename this issue.

draft-ietf-quic-qlog-quic-events.md
rmarx: I feel that loses a (potentially obvious, but still good to call out) aspect in how to use the `datagram_id` though... so maybe something like this then:

```suggestion
qlog defines its own mechanism for tracking coalescing across packet-level and
datagram-level qlog events, a "datagram identifier" carried in `datagram_id`
fields. qlog implementations that want to track coalescing can use this
mechanism, where multiple packet-level events sharing the same `datagram_id`
indicate they were coalesced in the same UDP datagram. The selection of specific
and locally-unique `datagram_id` values is an implementation choice. 
```

LPardue: Looks good

PR #407, opened 2024-03-04
Clarify UDP datagram namespace

### LINKED ISSUE #323: QUICDatagramsSent or UDPDatagramsSent
We currently have several events that can be used to indicate how QUIC packets are sent in UDP datagrams (`quic:datagrams_sent`, `quic:datagrams_received` etc.).

However, adding support for RFC9221 (Unreliable Datagram extension for QUIC), which adds the new DatagramFrame, causes some issues with the naming here.

For example, we also want a `datagram_data_moved` event in addition to the existing `(stream_)data_moved` event, but that would be ambiguous which "datagram" we're actually talking about (UDP datagrams or QUIC Datagram Frames). 

Since we've renamed the `transport` category to `quic` (before it was `transport:datagrams_*`), it makes sense to rename these events to `udp:datagrams_*` as well for consistency and clarity.

rmarx:
Closes #323.

I decided not to use the full `udp:` namespace since we might need to properly define that later (if qlog for UDP is ever done proper). So I now just added the prefix `udp_` to the necessary events to make clear these are UDP datagrams, not QUIC datagram frames.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -187,9 +187,9 @@ this specification.
 | quic:packet_dropped              | Base       | {{quic-packetdropped}} |
 | quic:packet_buffered             | Base       | {{quic-packetbuffered}} |
 | quic:packets_acked               | Extra      | {{quic-packetsacked}} |
-| quic:datagrams_sent              | Extra      | {{quic-datagramssent}} |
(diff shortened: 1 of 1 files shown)


LPardue (approved):
I think whatever we pick is an odd compromise but in the interest of getting things done, this rename is the shortest option that makes things clear without generating a lot of effort. :shipit:

PR #408, opened 2024-03-04
Clarify congestion_state_updated usage

### LINKED ISSUE #246: Should congestion_state_updated define some values for old and new?
congestion_state_updated fields include an `old` and `new` field that takes a `text` type. The prose that explains this event includes examples "slow_start, congestion_avoidance, application_limited, recovery"

Perhaps we should make these actual values in CDDL, add the BBR CC phases, and support general text too.

rmarx:
Closes #246.


LPardue (approved):
LGTM

PR #409, opened 2024-03-06
Clarify paramers_set merging of parameter values

### LINKED ISSUE #398: Difference between negotiated vs actually accepted/enabled TPs
Spinoff/splitoff from #369.

@hlandau  originally reported for `quic:parameters_set`:

I think this event needs more clarify in terms of the distinction between
attempting to negotiate a feature and it being enabled. It should be possible
to log the feature set a local or remote endpoint is requesting and it
should be possible to log the actually negotiated feature set. Possibly
this can be resolved via simple clarification. If a client connects to
a server, does it start by emitting parameters_set (owner=local) with (e.g.)
max_datagram_frame_size set, and then when the server responds without
it, emit parameters_set (owner=remote) without it? I guess this allows
the necessary information to be inferred.

If the absence of a field indicates it was not negotiated after receiving
transport parameters from a server, this creates a problem if this event has
to be used to log other parameters determined at different times (which seems
plausible for tls_cipher, early_data_enabled, etc.), as then logging
those parameters could be misinterpreted as a sign that the transport
parameters have been received and a feature has not been negotiated.
Either the cause of a parameters_set event should be clarified (maybe a trigger
field) or possibly this should be split into different event types.

rmarx:
Closes #398.

PTAL @hlandau. 

The suggestion of using `null` as a JSON value in #398 makes conceptual sense, but not really since we're working with CDDL to define the events and their fields, as it doesn't easily allow to make a type "nullable" for something like this. 

As such, I changed the recommendation to explicitly setting the default value if needed; I hope that's acceptable. 

For the rest, I mostly just used your text with minor editorial changes.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -635,10 +635,37 @@ Most of these settings are typically set once and never change. However, they
 are usually set at different times during the connection, so there will
 regularly be several instances of this event with different fields set.
 
-Note that some settings have two variations (one set locally, one requested by the
-remote peer). This is reflected in the `owner` field. As such, this field MUST be
-correct for all settings included a single event instance. If you need to log
-settings from two sides, you MUST emit two separate event instances.
+Some parameters have two variations: one set locally, and one requested by the
+remote peer, and both are valid and used during the connection. This is
+reflected in the `owner` field. As such, this field MUST be correct for all
+settings included a single event instance. If you need to log settings from two
+sides, you MUST emit two separate event instances.
+
+In contrast, some parameters are the result of negotiation between peers. In
+this case, one endpoint generally suggests a possible value for a parameter and
+its peer then suggests another possible value. The final value used by both
+endpoints is the result of some parameter-specific function applied over the two
+suggested values. Such negotiation SHOULD be expressed as a series of
+`parameters_set` events with the appropriate `owner` field value set. For
+example:
+
+- The local endpoint logs a `parameters_set` event with an `owner` of `local`
+  reflecting the parameter value(s) it would like to use.
+- The local endpoint receives parameter values from the peer and logs a
+  `parameters_set` event with an `owner` of `remote` reflecting the desired
+  value(s) expressed by the peer.
+- The local endpoint logs a `parameters_set` event with an `owner` of `local`
+  reflecting the resulting parameter value(s) determined from the suggested
+  parameter value(s) communicated by the local and remote endpoints.
+
+Consequently, each parameter value in a `parameters_set` event supersedes the
+previous value of that parameter if present. If a parameter does not appear in a
+given `parameters_set` event, its value is unchanged from the previous event
+which set that parameter. If a parameter never appears in a `parameters_set`
+event, it retains its default value. This default value thus cannot be restored
+implicitly (for example by omitting a parameter in a subsequent
+`parameters_set`) and, if needed, should instead be restored explicitly by
+re-assigning the default value in an new `parameters_set` event.
 
 In the case of connection resumption and 0-RTT, some of the server's parameters
 are stored up-front at the client and used for the initial connection startup.


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -635,10 +635,37 @@ Most of these settings are typically set once and never change. However, they
...
+  `parameters_set` event with an `owner` of `remote` reflecting the desired
+  value(s) expressed by the peer.
+- The local endpoint logs a `parameters_set` event with an `owner` of `local`
+  reflecting the resulting parameter value(s) determined from the suggested
+  parameter value(s) communicated by the local and remote endpoints.
+
+Consequently, each parameter value in a `parameters_set` event supersedes the
+previous value of that parameter if present. If a parameter does not appear in a
+given `parameters_set` event, its value is unchanged from the previous event
+which set that parameter. If a parameter never appears in a `parameters_set`
```
marten-seemann, line 664: This doesn't work for 0-RTT, does it?

rmarx: 0-RTT is special cased in the paragraph below this one and should use `parameters_restored`. (That's not to say that that is 100% watertight, I might have missed something in that flow, but it's indeed not the intent to use only `parameters_set` for 0-RTT and this new text is not intended for 0-RTT indeed).

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -635,10 +635,37 @@ Most of these settings are typically set once and never change. However, they
...
+suggested values. Such negotiation SHOULD be expressed as a series of
+`parameters_set` events with the appropriate `owner` field value set. For
+example:
+
+- The local endpoint logs a `parameters_set` event with an `owner` of `local`
+  reflecting the parameter value(s) it would like to use.
+- The local endpoint receives parameter values from the peer and logs a
+  `parameters_set` event with an `owner` of `remote` reflecting the desired
+  value(s) expressed by the peer.
+- The local endpoint logs a `parameters_set` event with an `owner` of `local`
```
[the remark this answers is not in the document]
hlandau: No objection to a separate owner value

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -635,10 +635,37 @@ Most of these settings are typically set once and never change. However, they
...
+  parameter value(s) communicated by the local and remote endpoints.
+
+Consequently, each parameter value in a `parameters_set` event supersedes the
+previous value of that parameter if present. If a parameter does not appear in a
+given `parameters_set` event, its value is unchanged from the previous event
+which set that parameter. If a parameter never appears in a `parameters_set`
+event, it retains its default value. This default value thus cannot be restored
+implicitly (for example by omitting a parameter in a subsequent
+`parameters_set`) and, if needed, should instead be restored explicitly by
+re-assigning the default value in an new `parameters_set` event.
```
hlandau, line 668: If we don't support `null` I guess that means some default values (i.e., absence of a TP) might be unrepresentable in qlog.

hlandau (approved):
LGTM, though one comment

draft-ietf-quic-qlog-quic-events.md
marten-seemann: I don't think this event should be logged more than once per connection. Transport parameters are only exchanged once, after all. If we have a separate owner value as suggested above, this property would also hold in qlog.

draft-ietf-quic-qlog-quic-events.md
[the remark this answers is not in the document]
marten-seemann: `negotiated`? I don't think this is super valuable to log, since it can easily be derived from the two `parameters_set` events anyway.

draft-ietf-quic-qlog-quic-events.md
rmarx: This event will be logged at least twice per connection by definition, since there are two sets of TPs: one for the local entity and one for the peer. This event also tracks more state than just transport parameters (e.g., TLS level params like resumption or early data) that are probably set in a different codepath/at a different time and benefit from having multiple `parameters_set`.

I agree with the subset of your statement that the event probably shouldn't be logged more than once per connection for specific fields with the same owner though. 

Still, for me personally the most important part here is `If a parameter does not appear in a given parameters_set event, its value is unchanged from the previous event which set that parameter.` to help make clear that subsequent `parameters_set` do not invalidate earlier values, which I agree with @hlandau is not 100% clear right now.

draft-ietf-quic-qlog-quic-events.md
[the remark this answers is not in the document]
rmarx: So, I'm still not 100% sure what we're proposing here...

Option A:
- log `parameters_set` with owner=local for local value
- log `parameters_set` with owner=remote for peer's value
- log `parameters_set` with owner=negotiated for final value

or Option B:
- log `parameters_set` with owner=negotiated for local value
- log `parameters_set` with owner=negotiated for peer's value
- log `parameters_set` with owner=negotiated for final value

or Option C:
- log `parameters_set` with owner=negotiated final value

I'm not sure I like C (lose quite some information there... don't know which values the peers proposed originally?).
I'm also not sure how either A or B improves upon the proposed text though... you're still logging 3 events?

draft-ietf-quic-qlog-quic-events.md
rmarx: I'm not sure I understand? Absence of a TP is simply that: absence of a TP in the `parameters_set`. 

Whether this means the TP was actually absent on the wire or whether the implementation didn't/couldn't log it is indeed unclear, but the qlogger can do what they want (e.g., also simply couldn't log `null`), so that's never fixable imo.

hlandau: I guess this is not so much a problem for TPs as they are only exchanged once. But it seems like this event is intended to be more general.

For other kinds of parameters which might be renegotiable I could see a problem if it is possible for a feature to transition from negotiated to the default/non-negotiated state.

Maybe it is not a problem in practice? I'm just envisaging this event type being expanded to other things besides TPs in future.

draft-ietf-quic-qlog-quic-events.md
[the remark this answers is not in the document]
rmarx: @marten-seemann : as discussed in the editor's meeting, PTAL and let me know which changes, if any, you'd like here.

rmarx:
Superseded by #493, which is now merged!

PR #412, opened 2024-03-16
Add raw bytes field support to CONNECTION_CLOSE.

### LINKED ISSUE #411: Logging non-UTF-8 reasons
Similarly to the ALPN string issue, we define the `CONNECTION_CLOSE` reason string here as text: https://github.com/quicwg/qlog/blob/57b679678b44d80234a0e10a08e53fe5ef9e7532/draft-ietf-quic-qlog-quic-events.md?plain=1#L313

The problem here is that RFC 9000 only says the reason string SHOULD be UTF-8. It could be arbitrary, or binary data, even if that is a bad idea.

It should be possible to log a reason string which an implementation sees which does not appear to be valid UTF-8.

LPardue:
Fixes #411

Same approach as the one we look for ALPNInformation. Allows endpoints to log either, because sometimes it is more useful to print and consume a string like "It's borked" than it's hex encoded equivalent.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -2051,6 +2051,11 @@ error. For known frame types, the appropriate string value is used. For unknown
 frame types, the numerical value without variable-length integer encoding is
 used.
 
+The CONNECTION_CLOSE reason is a byte sequences, that may be possible to
+present as UTF-8. The `reason_text` and `reason_bytes` fields provide support
+for either format. Implementations SHOULD log at least one format, but MAY log
+both or none.
+
 ~~~ cddl
 ErrorSpace = "transport" /
              "application"
@@ -2062,7 +2067,8 @@ ConnectionCloseFrame = {
                   CryptoError /
                   $ApplicationError /
                   uint64
-    ? reason: text
+    ? reason_text: text
+    ? reason_bytes: hexstring
 
     ; when error_space === "transport"
     ? trigger_frame_type: uint64 /
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -2052,7 +2052,7 @@ frame types, the numerical value without variable-length integer encoding is
 used.
 
 The CONNECTION_CLOSE reason is a byte sequences, that may be possible to
-present as UTF-8. The `reason_text` and `reason_bytes` fields provide support
+present as UTF-8. The `reason` and `reason_bytes` fields provide support
 for either format. Implementations SHOULD log at least one format, but MAY log
 both or none.
 
@@ -2067,7 +2067,7 @@ ConnectionCloseFrame = {
                   CryptoError /
                   $ApplicationError /
                   uint64
-    ? reason_text: text
+    ? reason: text
     ? reason_bytes: hexstring
 
     ; when error_space === "transport"
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -2051,10 +2051,11 @@ error. For known frame types, the appropriate string value is used. For unknown
 frame types, the numerical value without variable-length integer encoding is
 used.
 
-The CONNECTION_CLOSE reason is a byte sequences, that may be possible to
-present as UTF-8. The `reason` and `reason_bytes` fields provide support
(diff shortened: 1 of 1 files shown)


hlandau (approved):

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -2062,7 +2067,8 @@ ConnectionCloseFrame = {
                   CryptoError /
                   $ApplicationError /
                   uint64
-    ? reason: text
+    ? reason_text: text
```
marten-seemann, line 2070: I'd prefer to not rename this field.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -2051,6 +2051,11 @@ error. For known frame types, the appropriate string value is used. For unknown
 frame types, the numerical value without variable-length integer encoding is
 used.
 
+The CONNECTION_CLOSE reason is a byte sequences, that may be possible to
+present as UTF-8. The `reason_text` and `reason_bytes` fields provide support
```
marten-seemann, line 2055: That's not how I read RFC 9000. The field SHOULD be UTF-8, so the expectation is that it actually is. We're just creating an escape hatch here in case this assumption doesn't hold.

marten-seemann (changes requested):

draft-ietf-quic-qlog-quic-events.md
hlandau: Agreed - need better wording alignment

LPardue: If you're logging a reason sent by a peer you don't have any relationship with, there is no guarantee it is following the SHOULD. Therefore, it is only a possibility that the reason is presentable as UTF-8. I can't think of a more succint way to describe this possibility than the current text. If you have concrete suggestions for alternative text, please propose them :D

draft-ietf-quic-qlog-quic-events.md
LPardue: I can live with that

draft-ietf-quic-qlog-quic-events.md
LPardue: Spoke offline with Marten and came up with a slight variant

marten-seemann (approved):

rmarx:
I'm not the biggest fan of this not using the same pattern as we did for ALPNInformation... which would be more like this:

```
...
? reason: CloseReason
...

CloseReason = {
    ? byte_value: hexstring
    ? string_value: text
}
```

Is there a particular reason to change from that established precedent?

LPardue:
The main reason (pun intended) is that the `QUICALPNInformation` event contains 3 instances of ALPNInformation, which would make a flatter structure really hard to describe. Here, there is only one instance of the reason, so it seems of less value to create another structure type to hold it. 

I won't defend that choice with much vigor if others prefer a Structured approach.

rmarx:
I don't have a super-strong opinion on this, as this approach also makes sense. Especially seeing another need for this in https://github.com/quicwg/qlog/pull/414, we could reason this approach is the default, and ALPNInformation deviates from that for good reasons™

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -2051,6 +2051,12 @@ error. For known frame types, the appropriate string value is used. For unknown
 frame types, the numerical value without variable-length integer encoding is
 used.
 
+The CONNECTION_CLOSE reason phrase is a byte sequences. It is likely that this
+sequence is presentable as UTF-8, in which case it can be logged in the `reason`
+field. The `reason_bytes` field supports logging the bytes, which can be useful
```
rmarx, line 2056:
```suggestion
field. The `reason_bytes` field supports logging the raw bytes, which can be useful
```

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -2051,6 +2051,12 @@ error. For known frame types, the appropriate string value is used. For unknown
...
+when the value is not UTF-8 or when and endpoint does not want to decode it.
```
rmarx, line 2057:
```suggestion
when the value is not UTF-8 or when an endpoint does not want to decode it.
```

rmarx (changes requested):
small nits

rmarx (approved):

PR #414, opened 2024-03-17
Allow logging of header name and value bytes

### LINKED ISSUE #413: Logging of HTTP field names or values that contain illegal bytes
While normally HTTP header names and values are expected to be safe strings (possibly with even further restrictions on the chartacter set that is allowed), it's possible that an endpoint could receive garbage bytes that it would validate and reject. In such scenarios, it can be useful to log what was received for the purposes of debugging root causes.

The [H3HeadersFrame](https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html#name-h3headersframe) uses the `H3HTTPField` type, where names and values use the `text` type. This is unfortunate as it forces implementers into certain approaches to logging. For instance, Rust has a `f[rom_utf8_lossy()](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf8_lossy)` method that would insert a � for an invalid sequence. While safe, this discards information that could be useful to understand excatly what was received.

LPardue:
Fixes #413

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -542,15 +542,15 @@ HTTP/3 can transport field names or values that are not valid; see {{Section 4.2
 of RFC9114}}. While it is usually expected that fields can be presented as
 strings, where invalid values are received it can be useful to log the exact
 bytes. In order to support either format, the `H3HTTPField` has fields of either
-text or hexstring type. An `H3HTTPField` MUST include either the `name_text` or
+text or hexstring type. An `H3HTTPField` MUST include either the `name` or
 `name_bytes` field and MAY include both. An `H3HTTPField` MAY include a
-`value_text` or `value_bytes` field or neither.
+`value` or `value_bytes` field or neither.
 
 ~~~ cddl
 H3HTTPField = {
-    ? name_text: text
+    ? name: text
     ? name_bytes: hexstring
-    ? value_text: text
+    ? value: text
     ? value_bytes: hexstring
 }
 ~~~
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -560,7 +560,7 @@ headers: [
 ~~~
 {: #h3-headersframe-ex title="H3HeadersFrame example"}
 
-{{Section 4.2 of RFC9114}} and {{Section 5.1 of RFC 9110}} define rules for the
+{{Section 4.2 of RFC9114}} and {{Section 5.1 of RFC9110}} define rules for the
 characters used in HTTP field sections names and values. Characters outside the
 range are invalid and result in the message being treated as malformed.
 
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -530,7 +530,7 @@ H3HeadersFrame = {
 For example, the HTTP field section
 
 ~~~
-:path: value
+:path: /index.html
 :method: GET
 :authority: example.org
 :scheme: https
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -538,10 +538,20 @@ H3HeadersFrame = {
 ~~~
 {: #h3-headersframe-def title="H3HeadersFrame definition"}
 
+HTTP/3 can transport field names or values that are not valid; see {{Section 4.2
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-h3-events.md
```diff
@@ -505,10 +505,20 @@ H3HeadersFrame = {
 ~~~
 {: #h3-headersframe-def title="H3HeadersFrame definition"}
 
+HTTP/3 can transport field names or values that are not valid; see {{Section 4.2
```
marten-seemann, line 508:
```suggestion
HTTP/3 can encode field names or values that are not valid; see {{Section 4.2
```

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -505,10 +505,20 @@ H3HeadersFrame = {
...
+`value_text` or `value_bytes` field or neither.
+
 ~~~ cddl
 H3HTTPField = {
-    name: text
-    ? value: text
+    ? name_text: text
+    ? name_bytes: hexstring
+    ? value_text: text
+    ? value_bytes: hexstring
```
marten-seemann, line 521: Can we avoid renaming the existing fields?

draft-ietf-quic-qlog-h3-events.md
rmarx, line 508: Not valid what? Valid UTF-8? Valid semantically? valid in terms of range of bytes that are represented? etc.

draft-ietf-quic-qlog-h3-events.md
rmarx: If we decide on using the same approach here as in https://github.com/quicwg/qlog/pull/412, and there we call it `reason + reason_bytes` instead of `reason_text + reason_bytes`, then yeah, we should be consistent here as well and keep `name + name_bytes + value + value_bytes`

rmarx (changes requested):

draft-ietf-quic-qlog-h3-events.md
LPardue: I'll change to use the old names

rmarx:
It seems like we've firmly settled on this pattern for this type of thing:

```
field: string
field_bytes: hexstring
```

So I think all that's needed here is indeed changing `name_text` and `value_text` to `name` and `value` + my and Marten's nits (transport > encode, and "define valid" and should be good to go :)

LPardue:
I changed the text to make this all clearer (hopefully)

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -503,10 +505,38 @@ H3DataFrame = {
...
+    frame_type: "headers"
+    headers: [* H3HTTPField]
+}
+~~~
+{: #h3-headersframe-def title="H3HeadersFrame definition"}
+
+For example, the HTTP field section
+
+~~~
+:path: value
```
rmarx, line 533:
```suggestion
:path: /
```

for consistency with the example below (maybe even better to go for something like /my-resource ?

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -530,21 +560,17 @@ headers: [
...
-    name: text
-    ? value: text
-}
-~~~
-{: #h3field-def title="H3HTTPField definition"}
+{{Section 4.2 of RFC9114}} and {{Section 5.1 of RFC9110}} define rules for the
+characters used in HTTP field sections names and values. Characters outside the
+range are invalid and result in the message being treated as malformed.
+
+It is useful to log HTTP fields that are valid or invalid. Characters in the
```
rmarx, line 567:
```suggestion
Characters in the
```

This didn't flow well with the rest of the paragraph, so trying to group a bit more logically.

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -530,21 +560,17 @@ headers: [
...
-H3HTTPField = {
-    name: text
-    ? value: text
-}
-~~~
-{: #h3field-def title="H3HTTPField definition"}
+{{Section 4.2 of RFC9114}} and {{Section 5.1 of RFC9110}} define rules for the
+characters used in HTTP field sections names and values. Characters outside the
+range are invalid and result in the message being treated as malformed.
+
```
rmarx, line 566:
```suggestion
range are invalid and result in the message being treated as malformed. It can however be useful to also log these invalid HTTP fields.
```

rmarx (approved):
Mostly nits, so I'll go ahead and approve the general approach :)

PR #415, opened 2024-03-17
Define extension schema policy and registry

### LINKED ISSUE #283: Versioning for additional schema

LPardue:
Alternative to #284.

This follows Lennox's suggestion to take influence from RFC 8285 and reference
extension schema according to a URI. IETF documents can use URNs under a formal
namespace, and as such we register a new `qlog` URN Sub-namespace for
Registered Protocol Parameter Identifiers.

Fixes #283

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -95,7 +95,7 @@ implementation decision.
 
 # Overview
 
-This document describes how the QUIC protocol can be expressed in qlog. It
+This document describes how the HTTP/3 protocol can be expressed in qlog. It
 defines a qlog extension schema for use with the main schema {{QLOG-MAIN}}.
 HTTP/3 events are defined with a category, a name (the concatenation of
 "category" and "event"), an "importance", an optional "trigger", and "data"
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -50,9 +50,8 @@ informative:
 
 --- abstract
 
-This document describes concrete qlog event definitions and their metadata for
-HTTP/3-related events. These events can then be embedded in the higher
-level schema defined in {{QLOG-MAIN}}.
+This document defines a qlog extension schema containing concrete qlog event
+definitions and their metadata for HTTP/3-related events.
 
 --- note_Note_to_Readers
 
@@ -71,10 +70,12 @@ various programming languages can be found at
 
 # Introduction
 
-This document describes the values of the qlog name ("category" + "event") and
+This document defines a qlog extension schema ({{Section 9 of QLOG-MAIN}})
+containing concrete qlog event definitions and their metadata for HTTP/3-related
+events. Specifically, the values of the qlog name ("category" + "event") and
 "data" fields and their semantics for the HTTP/3 protocol {{RFC9114}} and some
-of extensions (see {{!EXTENDED-CONNECT=RFC9220}}, {{!H3_PRIORITIZATION=RFC9218}}
-and {{!H3-DATAGRAM=RFC9297}}).
+of it's extensions (see {{!EXTENDED-CONNECT=RFC9220}},
+{{!H3_PRIORITIZATION=RFC9218}} and {{!H3-DATAGRAM=RFC9297}}).
 
 ## Notational Conventions
 
@@ -94,10 +95,12 @@ implementation decision.
 
 # Overview
 
-This document describes how HTTP/3 can be expressed in qlog using the schema
-defined in {{QLOG-MAIN}}. HTTP/3 events are defined with a category, a name (the
-concatenation of "category" and "event"), an "importance", an optional
-"trigger", and "data" fields.
+This document defines a single event schema identified by the URI
+"urn:ietf:params:qlog:h3".
+
+HTTP/3 events are defined with a category, a name (the concatenation of
+"category" and "event"), an "importance", an optional "trigger", and "data"
+fields.
 
 Some data fields use complex datastructures. These are represented as enums or
 re-usable definitions, which are grouped together on the bottom of this document
@@ -105,6 +108,16 @@ for clarity.
 
 When any event from this document is included in a qlog trace, the
 "protocol_type" qlog array field MUST contain an entry with the value "HTTP3".
+
+## Draft Schema Identification
+
+Only implementations of the final, published RFC can use the schema identified
+by "urn:ietf:params:qlog:h3". Until such an RFC exists, implementations MUST NOT
+identify themselves using this URI.
+
+Implementations of draft versions of the extension schema MUST append the string
+"-" and the corresponding draft number to the URI. For example, draft 07 of this
+document is identified using the URI "urn:ietf:params:qlog:h3-07".
 
 ## Usage with QUIC
 
@@ -120,9 +133,9 @@ identifier, potentially suffixed by the vantagepoint type (For example,
 abcd1234_server.qlog would contain the server-side trace of the connection with
 GUID abcd1234).
 
-# HTTP/3 Event Overview
+# Event Overview
 
-This document defines events in two categories, written as lowercase to follow
+This document defines a single category of events, written as lowercase to follow
 convention: h3 ({{h3-ev}}).
 
 As described in {{Section 3.4.2 of QLOG-MAIN}}, the qlog "name" field is the
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -74,7 +74,7 @@ This document defines a qlog extension schema ({{Section 9 of QLOG-MAIN}})
 containing concrete qlog event definitions and their metadata for HTTP/3-related
 events. Specifically, the values of the qlog name ("category" + "event") and
 "data" fields and their semantics for the HTTP/3 protocol {{RFC9114}} and some
-of it's extensions (see {{!EXTENDED-CONNECT=RFC9220}},
+of its extensions (see {{!EXTENDED-CONNECT=RFC9220}},
 {{!H3_PRIORITIZATION=RFC9218}} and {{!H3-DATAGRAM=RFC9297}}).
 
 ## Notational Conventions
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -71,11 +71,11 @@ various programming languages can be found at
 # Introduction
 
 This document defines a qlog extension schema ({{Section 9 of QLOG-MAIN}})
-containing concrete qlog event definitions and their metadata for HTTP/3-related
-events. Specifically, the values of the qlog name ("category" + "event") and
-"data" fields and their semantics for the HTTP/3 protocol {{RFC9114}} and some
-of its extensions (see {{!EXTENDED-CONNECT=RFC9220}},
-{{!H3_PRIORITIZATION=RFC9218}} and {{!H3-DATAGRAM=RFC9297}}).
+containing concrete qlog event definitions related to HTTP/3. Specifically, the
+values of the qlog name ("category" + "event") and "data" fields and their
+semantics for the HTTP/3 protocol {{RFC9114}} and some of its extensions (see
+{{!EXTENDED-CONNECT=RFC9220}}, {{!H3_PRIORITIZATION=RFC9218}} and
+{{!H3-DATAGRAM=RFC9297}}).
 
 ## Notational Conventions
 
@@ -95,12 +95,12 @@ implementation decision.
 
 # Overview
 
-This document defines a single event schema identified by the URI
-"urn:ietf:params:qlog:h3".
-
+This document describes how the QUIC protocol can be expressed in qlog. It
+defines a qlog extension schema for use with the main schema {{QLOG-MAIN}}.
 HTTP/3 events are defined with a category, a name (the concatenation of
 "category" and "event"), an "importance", an optional "trigger", and "data"
-fields.
+fields.  The schema is identified by the URI "urn:ietf:params:qlog:h3", it
+contains a category identifier "h3".
 
 Some data fields use complex datastructures. These are represented as enums or
 re-usable definitions, which are grouped together on the bottom of this document
@@ -118,6 +118,8 @@ identify themselves using this URI.
 Implementations of draft versions of the extension schema MUST append the string
 "-" and the corresponding draft number to the URI. For example, draft 07 of this
 document is identified using the URI "urn:ietf:params:qlog:h3-07".
+
+The category name is not affected by this requirement.
 
 ## Usage with QUIC
 
@@ -676,6 +678,9 @@ identifer" registry.
 
 Extension URI:
 : urn:ietf:params:qlog:h3
+
+Category Identifier(s):
+: h3
 
 Description:
 : Event definitions related to the HTTP/3 application protocol.
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -658,7 +658,17 @@ document as well.
 
 # IANA Considerations
 
-There are no IANA considerations.
+This document registers a new entry in the "qlog event extension schema
+identifer" registry.
+
+Extension URI:
+: urn:ietf:params:qlog:h3
+
+Description:
+: Event definitions related to the HTTP/3 application protocol.
+
+Reference:
+: This Document.
 
 --- back
 
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1146,14 +1146,14 @@ SimulationMarker = {
 # Event extensibility {#event-extensibility}
 
 The main qlog schema defined by this document describes log and trace formats
-that have related events. A minimal set of common events are presented in
+that have related events. A minimal set of common events is presented in
 {{events}}. It is expected that logging is extended with specific, per-protocol
 event definitions that specify the name (category + type) and data needed for
 each individual event. Examples include the QUIC event definitions {{QLOG-QUIC}}
 and HTTP/3 event definitions {{QLOG-H3}}.
 
 New event definitions SHOULD follow the guidance in this section, which provides
-consistency that makes it easier for qlog implementers to support a wide variety
+consistency measures that make it easier for qlog implementers to support a wide variety
 of protocols.
 
 ## Extended Event Schema
@@ -1172,8 +1172,8 @@ Implementations that might record events from extension schemas SHOULD list all
 category identifiers in use. This is achieved by including the appropriate URI
 in the `additional_event_schemas` field of the QlogFile ({{qlog-file-schema}})
 or QlogFileSeq ({{qlog-file-seq-schema}}). The `additional_event_schema` is a
-hint to tools about the possible events that a qlog file might contain. The file
-may contain events that do not belong to a listed category identifier. Tools
+hint to tools about the possible event categories (and event types contained therein) that a qlog file might contain. The file
+may contain event types that do not belong to a listed category identifier. Tools
 MUST NOT treat this as an error; see {{tooling}}.
 
 Each extension schema is named by a URI. That URI MUST be absolute and MUST
@@ -1183,7 +1183,7 @@ after a "#" in the URI).
 For extension schema defined in RFCs, the URI used SHOULD be a URN starting with
 `urn:ietf:params:qlog:` followed by a registered, descriptive name.
 
-Private or non-standard extension schema a can use other URI formats. URIs that
+Private or non-standard extension schema can use other URI formats. URIs that
 contain a domain name SHOULD also contain a month-date in the form mmyyyy. The
 definition of the schema and assignment of the URI MUST have been authorized by
 the owner of the domain name on or very close to that date. (This avoids
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1159,34 +1159,51 @@ of protocols.
 ## Extended Event Schema
 
 New event definitions SHOULD be part of an extension schema, using the
-annotations and concepts presented in this section. Events can be grouped
-together into a category but cannot belong to multiple categories. A schema
-defines one or more qlog event categories.
+annotations and concepts presented in this section.
+
+A schema defines one or more qlog event categories that can contain multiple
+event definitions. Each event MUST only belong to a single category.
+
+It is not possible to extend a category with new events. Instead, extension
+schema MUST define new categories with globally unique identifiers. Per
+{{{#name-field}}}, event names are the concatenation of category and type.
+Therefore, unique category identifiers ensure events in qlog files are
+disambuiguated.
 
 Each extension schema is named by a URI. That URI MUST be absolute; it precisely
 identifies the format and meaning of the extension. URIs that contain a domain
 name SHOULD also contain a month-date in the form mmyyyy. The definition of the
 schema and assignment of the URI MUST have been authorized by the owner of the
 domain name on or very close to that date. (This avoids problems when domain
-names change ownership.) If the resource or document defines several categories,
-then the URI MUST identify the actual extension in use, e.g., using a fragment
-or query identifier (characters after a "#" or "?" in the URI).
+names change ownership.) If the resource or document defines several event
+categories, then the URI MUST identify the actual category in use, e.g., using
+a fragment or query identifier (characters after a "#" or "?" in the URI).
 
 For extensions defined in RFCs, the URI used SHOULD be a URN starting with
 `urn:ietf:params:qlog:` followed by a registered, descriptive name.
 
-A log file that that uses events from extension schema SHOULD list all schema
-identifiers in the `additional_event_schemas` field.
+A log file that uses events from extension schema SHOULD list all schema
+identifiers in the `additional_event_schemas` field; see {{qlog-file-schema}}
+and {{qlog-file-seq-schema}}.
 
-For example, a qlog file that uses two hypothetical standard qlog extension
-schema named "rick" and "morty", along with a private extension named "pickle"
-would indicate this as:
+In the following example, a qlog file contains two hypothetical standardized
+extension schema, along with a private extension. The first schema is named
+"rick" and consists of categories "roll", "astley", and "moranis". The second
+schema is named "john" and consists of categories "doe", "candy", and "goodman".
+The private schema is named "pickle" and consists of categories "pepper",
+"lilly", and "rick":
 
 ~~~
 additional_event_schemas = [
-                            urn:ietf:params:qlog:rick,
-                            urn:ietf:params:qlog:morty,
-                            https://example.com/032024/pickle.html#me
+                            urn:ietf:params:qlog:rick#roll,
+                            urn:ietf:params:qlog:rick#astley,
+                            urn:ietf:params:qlog:rick#moranis,
+                            urn:ietf:params:qlog:john#doe,
+                            urn:ietf:params:qlog:john#candy,
+                            urn:ietf:params:qlog:john#goodman,
+                            https://example.com/032024/pickle.html#pepper,
+                            https://example.com/032024/pickle.html#lilly,
+                            https://example.com/032024/pickle.html#rick,
                           ]
 ~~~
 
@@ -1691,23 +1708,29 @@ IANA is requested to create the "qlog event extension schema identifer" registry
 at [](https://www.iana.org/assignments/qlog) for the purpose of registering
 event extension schema. It has the following format:
 
-| Extension URI | Description | Reference |
-|||
+| Extension URI | Categor Identifier(s) | Description | Reference |
+||||
 
 No entries are registered by this document.
 
 The registry operates under the Expert Review policy, per {{Section 4.5 of
 !RFC8126}}.  When reviewing requests, the expert SHOULD check that the guidance
 in {{event-extensibility}} and in {{privacy}} has been duly considered. However,
-the outcome of this check cannot be used as a basis for rejection. The expert
-SHOULD also check that the request URI is appropriate to the extension and
+the outcome of this check cannot be used as a basis for rejection.
+
+The expert SHOULD check that the request URI is appropriate to the extension and
 unique, including determining if the number of requested URIs are proportional
-to the number of extension points.
+to the number of extension points.  The expert MUST check that category
+identifiers are unique. Non-unique URI or category names are sufficent grounds
+for rejection.
 
 Registration requests should use the following template:
 
 Extension URI:
-: \[The extension identifier\]
+: \[the extension schema identifier\]
+
+Category identifiers:
+: \[a comma-seperated list of the full set of categories belonging to the schema\]
 
 Description:
 : \[a description of the extension schema\]
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1178,7 +1178,7 @@ For extensions defined in RFCs, the URI used SHOULD be a URN starting with
 A log file that that uses events from extension schema SHOULD list all schema
 identifiers in the `additional_event_schemas` field.
 
-For example, a qlog file that uses three hypothetical standard qlog extension
+For example, a qlog file that uses two hypothetical standard qlog extension
 schema named "rick" and "morty", along with a private extension named "pickle"
 would indicate this as:
 
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1695,13 +1695,13 @@ event extension schema. It has the following format:
 
 No entries are registered by this document.
 
-The registry operates under the Expert Review policy, per {{Section 4.5 of !RFC8126}}.  When reviewing requests, the expert SHOULD check that:
-
-: The guidance in {{event-extensibility}} and in {{privacy}} has been duly
-considered. However, the outcome of this check cannot be used as a basis for
-rejection.
-
-: The request URI is appropriate to the extension and unique.
+The registry operates under the Expert Review policy, per {{Section 4.5 of
+!RFC8126}}.  When reviewing requests, the expert SHOULD check that the guidance
+in {{event-extensibility}} and in {{privacy}} has been duly considered. However,
+the outcome of this check cannot be used as a basis for rejection. The expert
+SHOULD also check that the request URI is appropriate to the extension and
+unique, including determining if the number of requested URIs are proportional
+to the number of extension points.
 
 Registration requests should use the following template:
 
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1723,7 +1723,8 @@ Universities.
 
 Thanks to Jana Iyengar, Brian Trammell, Dmitri Tikhonov, Stephen Petrides, Jari
 Arkko, Marcus Ihlar, Victor Vasiliev, Mirja Kühlewind, Jeremy Lainé, Kazu
-Yamamoto, Christian Huitema and Hugo Landau for their feedback and suggestions.
+Yamamoto, Christian Huitema, Hugo Landau, and Jonathan Lennow for their feedback
+and suggestions.
 
 # Change Log
 {:numbered="false" removeinrfc="true"}
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1170,8 +1170,7 @@ schemas MUST define new categories with globally unique identifiers. Per
 Therefore, unique category identifiers ensure events in qlog files are
 disambuiguated.
 
-Each extension schema is named by a URI. That URI MUST be absolute; it precisely
-identifies the format and meaning of the extension. URIs that contain a domain
+Each extension schema is named by a URI. That URI MUST be absolute. URIs that contain a domain
 name SHOULD also contain a month-date in the form mmyyyy. The definition of the
 schema and assignment of the URI MUST have been authorized by the owner of the
 domain name on or very close to that date. (This avoids problems when domain
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1176,8 +1176,8 @@ name SHOULD also contain a month-date in the form mmyyyy. The definition of the
 schema and assignment of the URI MUST have been authorized by the owner of the
 domain name on or very close to that date. (This avoids problems when domain
 names change ownership.) If the extension schema contains several event
-categories, then the URI MUST identify the actual category in use, e.g., using
-a fragment or query identifier (characters after a "#" or "?" in the URI).
+categories, then the URI MUST identify the actual category in use using
+a fragment identifier (characters after a "#" in the URI).
 
 For extensions defined in RFCs, the URI used SHOULD be a URN starting with
 `urn:ietf:params:qlog:` followed by a registered, descriptive name.
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1166,7 +1166,7 @@ defines one or more qlog event categories.
 Each extension schema is named by a URI. That URI MUST be absolute; it precisely
 identifies the format and meaning of the extension. URIs that contain a domain
 name SHOULD also contain a month-date in the form mmyyyy. The definition of the
-element and assignment of the URI MUST have been authorized by the owner of the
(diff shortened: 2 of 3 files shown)


LPardue:
@JonathanLennox I tried to borrow a lot from RFC 8285 but not take it fully verbatim. Would welcome your thoughts.

JonathanLennox:
This matches my suggestion, looks good to me!

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,52 @@ SimulationMarker = {
...
+
+## Extended Event Schema
+
+New events SHOULD be defined using an extension schema, using the annotations
+and concepts presented in this section.
+
+Each extension schema is named by a URI. That URI MUST be absolute; it precisely
+identifies the format and meaning of the extension. URIs that contain a domain
+name SHOULD also contain a month-date in the form mmyyyy. The definition of the
+element and assignment of the URI MUST have been authorized by the owner of the
```
rmarx, line 1167: unclear what "element" is referring to specifically here...

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,52 @@ SimulationMarker = {
...
+domain name on or very close to that date. (This avoids problems when domain
+names change ownership.) If the resource or document defines several categories
+or groups of events, then the URI MUST identify the actual extension in use,
```
rmarx, line 1170: this could do with an example maybe? not sure what that would look like if you're for example using quic events... would you need to list #connectivity and #transport as separate subentries in `additional_event_schema` or not? Where do you draw the line?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,52 @@ SimulationMarker = {
...
+URI).
 
-This document defines the main schema for the qlog format together with some
-common events, which on their own do not provide much logging utility. It is
-expected that logging is extended with specific, per-protocol event definitions
-that specify the name (category + type) and data needed for each individual
-event. Examples include the QUIC event definitions {{QLOG-QUIC}} and HTTP/3
-event definitions {{QLOG-H3}}.
+For extensions defined in RFCs, the URI used SHOULD be a URN starting with
+`urn:ietf:params:qlog` followed by a registered, descriptive name.
```
rmarx, line 1175:
```suggestion
`urn:ietf:params:qlog:` followed by a registered, descriptive name.
```

Prefix should end with `:` no?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,52 @@ SimulationMarker = {
...
+identifiers in the `additional_event_schema` field.
+
+For example, a qlog file that uses two hypothetical qlog extension schema named
+"rick" and "morty" would indicate this as:
+
+~~~
+additional_event_schema = [
+                            urn:ietf:params:qlog:rick,
+                            urn:ietf:params:qlog:morty
+                          ]
```
rmarx, line 1187: This could do with an extension of the example to also show at least one domainname-based extension (i.e., things that are not defined in RFCs), with the mmyyyy date in the URI

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1643,7 +1723,8 @@ Universities.
 
 Thanks to Jana Iyengar, Brian Trammell, Dmitri Tikhonov, Stephen Petrides, Jari
 Arkko, Marcus Ihlar, Victor Vasiliev, Mirja Kühlewind, Jeremy Lainé, Kazu
-Yamamoto, Christian Huitema and Hugo Landau for their feedback and suggestions.
+Yamamoto, Christian Huitema, Hugo Landau, and Jonathan Lennow for their feedback
```
rmarx, line 1726:
```suggestion
Yamamoto, Christian Huitema, Hugo Landau, and Jonathan Lennox for their feedback
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,52 @@ SimulationMarker = {
...
+each individual event. Examples include the QUIC event definitions {{QLOG-QUIC}}
+and HTTP/3 event definitions {{QLOG-H3}}.
+
+New event definitions SHOULD follow the guidance in this section, which provides
+consistency that makes it easier for qlog implementers to support a wide variety
+of protocols.
+
+## Extended Event Schema
+
+New events SHOULD be defined using an extension schema, using the annotations
```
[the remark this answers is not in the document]
LPardue: Thought a bit more on this. What I think we want is for logical groups of qlog events to be defined under a single schema. There should not be a restriction that a single IETF document defines only a single schema - for example, we might want to have a single document that defines events for a bunch of QUIC extensions and split them out by extension-specific schemas.

I don't think the current text, or the proposal, articulate this well. Will try to come up with a some different text.

draft-ietf-quic-qlog-main-schema.md
LPardue: nor me, I copy-pasted it from RFC 8285. Will change.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -215,6 +215,7 @@ component traces, defined in {{qlog-file-def}} as:
 QlogFile = {
     qlog_version: text
     ? qlog_format: text .default "JSON"
+    ? additional_event_schema: [+ text]
```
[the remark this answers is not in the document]
LPardue: `schemas` seems most correct, will fix.

While we are here, I wonder if CDDL has a URI type?

draft-ietf-quic-qlog-main-schema.md
rmarx: Is there an option to say: if you don't define # or ?, then you mean EVERYTHING included in the URI. If you want to select only specific parts, use # or ? to specify that? 

e.g., urn:ietf:params:qlog:quic means ALL QUIC events, while urn:ietf:params:qlog:quic#recovery means ONLY the recovery events and nothing else? How does RFC 8285 manage such things?

draft-ietf-quic-qlog-main-schema.md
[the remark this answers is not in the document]
LPardue: Instead of `text` is there a `uri` or somthinh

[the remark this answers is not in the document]
LPardue: Thanks for the clarification

draft-ietf-quic-qlog-main-schema.md
LPardue: This is an open question we should ask in the meeting

LPardue:
I pushed some changes to update the text in h3, if we like that approach I will carry it over to the quic doc.

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -71,10 +70,12 @@ various programming languages can be found at
...
 # Introduction
 
-This document describes the values of the qlog name ("category" + "event") and
+This document defines a qlog extension schema ({{Section 9 of QLOG-MAIN}})
+containing concrete qlog event definitions and their metadata for HTTP/3-related
+events. Specifically, the values of the qlog name ("category" + "event") and
 "data" fields and their semantics for the HTTP/3 protocol {{RFC9114}} and some
-of extensions (see {{!EXTENDED-CONNECT=RFC9220}}, {{!H3_PRIORITIZATION=RFC9218}}
-and {{!H3-DATAGRAM=RFC9297}}).
+of it's extensions (see {{!EXTENDED-CONNECT=RFC9220}},
```
rmarx, line 77:
```suggestion
of its extensions (see {{!EXTENDED-CONNECT=RFC9220}},
```

draft-ietf-quic-qlog-main-schema.md
LPardue: I spoke with some people in the hallways. There's a reason that RFC 8285 does it that way, due to how RTP stuff works. However, everyone seemed to agree that requiring the full expansions wasn't that onerous and was more clear. Allowing the url without those seems like a microoptimization.

There's another reason the full expansion is nice. If any documents in the future extend a namespace, then its not clear what extensions are used or not. That makes the implicit include quite insipid.

LPardue:
On thinking some more, I think allowing extension of categories with new events is a complication we can avoid. Specifically, allowing different documents to extend a category risks non-unique event names if the document author is not careful with type names. For example. if `ext-schema-a` and `ext-schema-b` both extend `h3` with a `my-ev-type` event (concating to the name `h3:my-ev-type`, then they'll be fine until someone tries to use both. It's going to be very hard to police event type names, especially for private extensions.

Instead, we can enforce that new schema need to define new globally-unique categories. For example, if we wanted to add new quic events in the future as part of a rollup extension schema, we could define `quic-2030`, `recovery-2030` etc. Adding this to the guidance pushes private extensions to avoid the problem, and requires standard extensions to do it.

I pushed an update that attempts to explain all this.

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -658,7 +673,20 @@ document as well.
 
 # IANA Considerations
 
-There are no IANA considerations.
+This document registers a new entry in the "qlog event extension schema
+identifer" registry.
+
+Extension URI:
```
rmarx, line 679: Not sure if this is actually an issue, but IIUC, this would mean that listing the H3 events in `additional_event_schemas` will lead to a URI like this:

`urn:ietf:params:qlog:h3#h3` 

Conceptually fine, but the duplication of h3 is a bit weird. Makes me wonder if in general we should go for the "full acronym" in the extension URI (i.e., http3 instead of h3, analogous to quic instead of q1)?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+New event definitions SHOULD follow the guidance in this section, which provides
+consistency that makes it easier for qlog implementers to support a wide variety
+of protocols.
+
+## Extended Event Schema
+
+New event definitions SHOULD be part of an extension schema, using the
+annotations and concepts presented in this section.
+
+A schema defines one or more qlog event categories that can contain multiple
```
rmarx, line 1164:
```suggestion
An extension schema defines one or more qlog event categories that can contain multiple
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+event definitions. Each event MUST only belong to a single category.
 
-This document defines the main schema for the qlog format together with some
-common events, which on their own do not provide much logging utility. It is
-expected that logging is extended with specific, per-protocol event definitions
-that specify the name (category + type) and data needed for each individual
-event. Examples include the QUIC event definitions {{QLOG-QUIC}} and HTTP/3
-event definitions {{QLOG-H3}}.
+It is not possible to extend a category with new events. Instead, extension
```
rmarx, line 1167:
```suggestion
It is not possible to extend an existing category, defined in a different document, with new events. Instead, extension
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+schema MUST define new categories with globally unique identifiers. Per
```
rmarx, line 1168:
```suggestion
schemas MUST define new categories with globally unique identifiers. Per
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+{{{#name-field}}}, event names are the concatenation of category and type.
+Therefore, unique category identifiers ensure events in qlog files are
+disambuiguated.
+
+Each extension schema is named by a URI. That URI MUST be absolute; it precisely
+identifies the format and meaning of the extension. URIs that contain a domain
```
rmarx, line 1174: Not exactly sure what the intention is of including "format" here... might be confused with "serialization format", which is certainly not the intent. Replace with a different word or remove?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+name SHOULD also contain a month-date in the form mmyyyy. The definition of the
+schema and assignment of the URI MUST have been authorized by the owner of the
+domain name on or very close to that date. (This avoids problems when domain
+names change ownership.) If the resource or document defines several event
```
rmarx, line 1178:
```suggestion
names change ownership.) If the extension schema contains several event
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+categories, then the URI MUST identify the actual category in use, e.g., using
+a fragment or query identifier (characters after a "#" or "?" in the URI).
```
rmarx, line 1180: I wonder why we'd use an "e.g.," here... are there other accepted options? if not, remove the e.g.,. If yes, I'd like to see at least one in the example :)

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+
+For extensions defined in RFCs, the URI used SHOULD be a URN starting with
+`urn:ietf:params:qlog:` followed by a registered, descriptive name.
+
+A log file that uses events from extension schema SHOULD list all schema
```
rmarx, line 1185:
```suggestion
A log file that uses events from extension schemas SHOULD list all extension schema
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+identifiers in the `additional_event_schemas` field; see {{qlog-file-schema}}
```
rmarx, line 1186:
```suggestion
identifiers (potentially multiple per extension schema in the case multiple categories are used) in the `additional_event_schemas` field; see {{qlog-file-schema}}
```

I feel that at this point, there's lots of ambiguity between a "schema identifier" and a specification of it for a specific category and what should actually be listed in the `additional_event_schemas`. The example below makes it a bit clearer, but I think we need more in this prose as well. I think the above is the minimum we could get away with, but potentially more text is needed (and maybe a new term like "category identifier"?)

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+and {{qlog-file-seq-schema}}.
+
+In the following example, a qlog file contains two hypothetical standardized
```
rmarx, line 1189:
```suggestion
In the following example, a qlog file contains events defined in two hypothetical standardized
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+                            urn:ietf:params:qlog:rick#roll,
+                            urn:ietf:params:qlog:rick#astley,
+                            urn:ietf:params:qlog:rick#moranis,
+                            urn:ietf:params:qlog:john#doe,
+                            urn:ietf:params:qlog:john#candy,
+                            urn:ietf:params:qlog:john#goodman,
+                            https://example.com/032024/pickle.html#pepper,
+                            https://example.com/032024/pickle.html#lilly,
+                            https://example.com/032024/pickle.html#rick,
+                          ]
```
rmarx, line 1207:
```suggestion
"additional_event_schemas": [
                            "urn:ietf:params:qlog:rick#roll",
                            "urn:ietf:params:qlog:rick#astley",
                            "urn:ietf:params:qlog:rick#moranis",
                            "urn:ietf:params:qlog:john#doe",
                            "urn:ietf:params:qlog:john#candy",
                            "urn:ietf:params:qlog:john#goodman",
                            "https://example.com/032024/pickle.html#pepper",
                            "https://example.com/032024/pickle.html#lilly",
                            "https://example.com/032024/pickle.html#rick",
                          ]
```

Since it's an example of a JSON serialized qlog file, it should be valid JSON.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+~~~
```
rmarx, line 1208:
```suggestion
~~~
{: #additional-event-schema-ex title="Example of using the additional-event-schema field"}
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1629,9 +1689,54 @@ The most sensitive data in qlog is typically contained in RawInfo type fields
...
+
+IANA Registry Reference:
+: [](https://www.iana.org/assignments/qlog){: brackets="angle"}
+
+
+IANA is requested to create the "qlog event extension schema identifer" registry
+at [](https://www.iana.org/assignments/qlog) for the purpose of registering
+event extension schema. It has the following format:
+
+| Extension URI | Categor Identifier(s) | Description | Reference |
```
rmarx, line 1711:
```suggestion
| Extension URI | Category Identifier(s) | Description | Reference |
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,72 @@ SimulationMarker = {
...
+names change ownership.) If the resource or document defines several event
+categories, then the URI MUST identify the actual category in use, e.g., using
+a fragment or query identifier (characters after a "#" or "?" in the URI).
+
+For extensions defined in RFCs, the URI used SHOULD be a URN starting with
+`urn:ietf:params:qlog:` followed by a registered, descriptive name.
+
+A log file that uses events from extension schema SHOULD list all schema
+identifiers in the `additional_event_schemas` field; see {{qlog-file-schema}}
+and {{qlog-file-seq-schema}}.
```
rmarx, line 1187: So here, I'm wondering if we should give guidance to tools on how to actually use the `additional_event_schemas` field? Like, they SHOULD use it to see if they support the listed event categories, but they SHOULD NOT bork on unexpected events (and rather just ignore them). 

put differently: I think we need to make clear that tools/users MUST NOT treat this as a fully qualified list and expect all/none events to be present, but rather more as a hint to further improve operations.

rmarx (changes requested):
As you'll see, a few remarks, most tiny, but a few that imo should be tackled.

I'm personally not a big fan of disallowing extensions to extend existing protocol/category namespaces... I get the rationale, but I think this will cause unnecessary churn and overhead for implementers (listing all extension docs in `additional_event_schema`, requesting addition to IANA, tools having to decide how to validate etc.). That said... I don't really have a better solution atm to the issues this solves :( 

If we do keep this, imo we do need to add some guidance on naming of new extensions for existing qlog definitions.

For example, https://www.ietf.org/archive/id/draft-ietf-tsvwg-careful-resume-07.html#name-cr_phase-event, is really a good fit for "quic:recovery", but wouldn't be able to just use that... so what should it do? 

It can't just do `quic-ext:recovery` because that would pollute `quic-ext` for all other extensions... it can't do `quic-rfcXYZA:recovery` because juck. It might do `quic-careful-resume:recovery` but this should probably be recommended by the main spec to prevent a weird growth of these things going forward (UNLESS you expect all this to be caught in the IANA registration process, where expert reviews can catch `quic-ext` and require it be renamed to something like `quic-careful-resume`? even then though... a bit of prose with an example could be nice here).

draft-ietf-quic-qlog-h3-events.md
LPardue: yeah that makes sense

draft-ietf-quic-qlog-main-schema.md
LPardue: I don't think we need the "different document" qualification, there may be no document at all for something like a private extension. The blanket rule is always the same, no extension.

draft-ietf-quic-qlog-main-schema.md
LPardue: lol, still a hangover from copy/paste. What we should say here is that a schema URI precisely identifies the category name and the event type(s) (the CDDL definition)

draft-ietf-quic-qlog-main-schema.md
LPardue: This was also a copy/paste. I don't see a good argument for allowing flexibility to be honest. How about we just restrict it to `#`?

draft-ietf-quic-qlog-main-schema.md
LPardue: Yeah that would be good to have.

LPardue:
> As you'll see, a few remarks, most tiny, but a few that imo should be tackled.
> 
> I'm personally not a big fan of disallowing extensions to extend existing protocol/category namespaces... I get the rationale, but I think this will cause unnecessary churn and overhead for implementers (listing all extension docs in `additional_event_schema`, requesting addition to IANA, tools having to decide how to validate etc.). That said... I don't really have a better solution atm to the issues this solves :(

I don't think it's that much extra work really. An IANA registration request is as simple as an email that will be forwarded to the designated experts (probably you, me and anyone else that wants to volunteer). Lazy implementations can omit the additional schema field at the cost of risk that tools will break. 

If we allowed extension to categories while only recommending that `additional_event_schema` is provided, we risk tools having to deal with ambiguous types, which sucks. 

> 
> If we do keep this, imo we do need to add some guidance on naming of new extensions for existing qlog definitions.
> 
> For example, https://www.ietf.org/archive/id/draft-ietf-tsvwg-careful-resume-07.html#name-cr_phase-event, is really a good fit for "quic:recovery", but wouldn't be able to just use that... so what should it do?
> 
> It can't just do `quic-ext:recovery` because that would pollute `quic-ext` for all other extensions... it can't do `quic-rfcXYZA:recovery` because juck. It might do `quic-careful-resume:recovery` but this should probably be recommended by the main spec to prevent a weird growth of these things going forward (UNLESS you expect all this to be caught in the IANA registration process, where expert reviews can catch `quic-ext` and require it be renamed to something like `quic-careful-resume`? even then though... a bit of prose with an example could be nice here).

The careful resume event is a good example but I disagree about your analysis. The definition of the event leaves it completely ambiguous about what category it belongs to, so it is unclear what it's serialized name would be. You seem to assume the event is for QUIC, but careful resume is a transport-independent.  So, assuming we get this PR hammered out, the careful resume document would have clearer guidance to follow about defining an extension schema. My recommendation would for it to define a new URI like `urn:ietf:params:qlog:careful-resume` and category identifier `cr`, then the event type could be `phase_updated` and it's serialized name would be `cr:phase_updated`.

rmarx:
> If we allowed extension to categories while only recommending that additional_event_schema is provided, we risk tools having to deal with ambiguous types, which sucks.

I agree, and that's why I said I also don't see a better solution (sadly) :) 

> The careful resume event is a good example but I disagree about your analysis

So ok, maybe careful-resume can get away with its own identifier outside of quic :) but then let's assume a loss-bits extension that is specific to QUIC's header and defines the `loss_bits_updated` event. Then it's still a bit unclear if they should have `urn:ietf:params:qlog:loss-bits` or `urn:ietf:params:qlog:quic-loss-bits` or `urn:ietf:params:qlog:quic-ext-loss-bits` or... (or if this even matters as much ;)) and thus if we should have a bit of guidance for that (i.e., extensions of existing protocols)

draft-ietf-quic-qlog-h3-events.md
LPardue: how about `urn:ietf:params:qlog:http#h3`? That would make it easy to add support for other HTTP versions or extensions as new catagories

draft-ietf-quic-qlog-main-schema.md
LPardue: done

draft-ietf-quic-qlog-h3-events.md
LPardue: Ah no my suggestion doesn't work because we don't allow category extensions

LPardue: Actually that's different. So its an open question, should we allow new categories to be added to a schema? 

In other words if the h3 spec registered `urn:ietf:params:qlog:http` and category `h3`, should we let other documents add categories? I think it is a reasonable thing to do but if so we would need to tweak the registration guidance.

draft-ietf-quic-qlog-main-schema.md
LPardue: See new text

draft-ietf-quic-qlog-main-schema.md
LPardue: See new text

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
 ~~~
 {: #simulation-marker-def title="SimulationMarker definition"}
 
-# Event definition guidelines
+# Event extensibility {#event-extensibility}
+
+The main qlog schema defined by this document describes log and trace formats
+that have related events. A minimal set of common events are presented in
```
rmarx, line 1149:
```suggestion
that have related events. A minimal set of common events is presented in
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+{{events}}. It is expected that logging is extended with specific, per-protocol
+event definitions that specify the name (category + type) and data needed for
+each individual event. Examples include the QUIC event definitions {{QLOG-QUIC}}
+and HTTP/3 event definitions {{QLOG-H3}}.
+
+New event definitions SHOULD follow the guidance in this section, which provides
+consistency that makes it easier for qlog implementers to support a wide variety
```
rmarx, line 1156:
```suggestion
consistency measures that make it easier for qlog implementers to support a wide variety
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+of protocols.
+
+## Extended Event Schema
+
+New event definitions SHOULD be part of an extension schema, using the
+annotations and concepts presented in this section.
+
+An extension schema defines one or more qlog event categories that can contain
+multiple event definitions. Each category MUST have a globally unique category
```
rmarx, line 1165: we are being ambiguous as to what exactly an "event definition" is.

Above, you define it as `per-protocol
event definitions that specify the name (category + type) and data` while here, it's more correct to say each category contains multiple event types... in the next sentence then, you switch to just the word "event".

It's not a major issue, but I feel we need consistent naming for these things...

For that, I would say `event type` here instead of `event definition` and in the next sentence `Each event type MUST only belong` instead of `Each event MUST`


Related: I think it would really help to have an illustrative example of this at the start of this part

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+identifier. Each event MUST only belong to a single category. Per
+{{{#name-field}}}, event names are the concatenation of category and type.
+Therefore, these requirements ensure there are no ambiguities when multiple
+extension schema are in use.
```
rmarx, line 1169: I understand what you're saying because I have context, and it's an important thing to highlight, BUT it's at the wrong place in the text... new readers don't know why this is important because you haven't talked about the URIs yet (and how event names tie into those).

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+
+Implementations that might record events from extension schemas SHOULD list all
+category identifiers in use. This is achieved by including the appropriate URI
+in the `additional_event_schemas` field of the QlogFile ({{qlog-file-schema}})
+or QlogFileSeq ({{qlog-file-seq-schema}}). The `additional_event_schema` is a
+hint to tools about the possible events that a qlog file might contain. The file
```
rmarx, line 1175:
```suggestion
hint to tools about the possible event categories (and event types contained therein) that a qlog file might contain. The file
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+may contain events that do not belong to a listed category identifier. Tools
```
rmarx, line 1176:
```suggestion
may contain event types that do not belong to a listed category identifier. Tools
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+MUST NOT treat this as an error; see {{tooling}}.
+
+Each extension schema is named by a URI. That URI MUST be absolute and MUST
+include a category identifier, indicated using a fragment identifier (characters
```
rmarx, line 1180: This is inconsistent with how you defined a "schema" at the start. There you say:

> An extension schema defines one or more qlog event categories

and here 1 schema is named by 1 URI, but that URI then somehow contains a category identifier? So the URI doesn't identify the schema, it instead identifies a single category WITHIN an extension schema.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+after a "#" in the URI).
+
+For extension schema defined in RFCs, the URI used SHOULD be a URN starting with
+`urn:ietf:params:qlog:` followed by a registered, descriptive name.
+
+Private or non-standard extension schema a can use other URI formats. URIs that
```
rmarx, line 1186:
```suggestion
Private or non-standard extension schema can use other URI formats. URIs that
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+definition of the schema and assignment of the URI MUST have been authorized by
+the owner of the domain name on or very close to that date. (This avoids
+problems when domain names change ownership.)
+
+In the following example, a qlog file contains events defined in two hypothetical standardized
+extension schema, along with a private extension. The first schema is named
+"rick" and consists of categories "roll", "astley", and "moranis". The second
+schema is named "john" and consists of categories "doe", "candy", and "goodman".
+The private schema is named "pickle" and consists of categories "pepper",
+"lilly", and "rick":
```
rmarx, line 1197: As I said above, I think it would REALLY help to have this example (or part of it) at the start of this section, as it makes it much clearer what you mean by schema and category

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1137,18 +1143,76 @@ SimulationMarker = {
...
+`urn:ietf:params:qlog:` followed by a registered, descriptive name.
+
+Private or non-standard extension schema a can use other URI formats. URIs that
+contain a domain name SHOULD also contain a month-date in the form mmyyyy. The
+definition of the schema and assignment of the URI MUST have been authorized by
+the owner of the domain name on or very close to that date. (This avoids
+problems when domain names change ownership.)
+
+In the following example, a qlog file contains events defined in two hypothetical standardized
+extension schema, along with a private extension. The first schema is named
```
rmarx, line 1193: it's a bit confusing that here we have 3 schema in a single file... while not impossible, most files will have a single schema (i.e., quic and h3).

I would rework this to saying that we have a single file with 2 schema (rick and john) and another file with a private extension schema (pickle).

rmarx (changes requested):
In general, the direction remains ok, but I think this could do with (a lot) more consistency and clarification... 

I think having a concrete example at the start of the section that can be referred back to by later paragraphs would be a big help (even though that's not super-common in IETF docs, I know). Even if you don't want to move the current example up ,we need an example that clearly indicates what's a schema, a category, an event type (and possibly what's an event identifier, and an "event", if you want to keep using those terms). 
 
I also still have some reservations on (dis)allowing extension of existing categories and schema's, but will post this in a separate comment below.

rmarx:
As mentioned above, I remain on the edge on what exactly we would allow to be extended by new documents and why...


Recent comments from Lucas say things to the effect of

> That does raise the question whether we should allow adding new categories to existing schema. That seems something that is OK.
    
but [earlier he said](https://github.com/quicwg/qlog/pull/415/#issuecomment-2029624210) it's not ok to add new events to existing categories, because you get ambiguous event names (category + type)
    
 > Specifically, allowing different documents to extend a category risks non-unique event names if the document author is not careful with type names

So there's an important nuance here: **extending an existing _category_ in an existing schema with new event types is NOT the same as extending an existing _schema_ with a new category and new events types**
        
 For example:

        urn:ietf:params:qlog:http is the DOC/schema
        urn:ietf:params:qlog:http#h3 is the category
        urn:ietf:params:qlog:http#h3:frame-created is the type
        
If you allow new docs to re-use the h3 category:
        
        Document L defines:
        urn:ietf:params:qlog:http#h3:new-event-Z

        Document K also defines:
        urn:ietf:params:qlog:http#h3:new-event-Z
        
Then you indeed get collisions / ambiguities (because both would use the URI urn:ietf:params:qlog:http#h3 in `additional_event_schemas`). So far, so good, and I agree that's a problem.

HOWEVER, I don't really see how only re-using the schema itself would (completely) prevent that problem, _except for IANA enforcing uniqueness._ 

If you allow new docs to only re-use the http schema and not the h3 category:

        Document A defines:
        urn:ietf:params:qlog:http#h4:new-event-X
        urn:ietf:params:qlog:http#h4:new-event-Z

        Document B also defines:
        urn:ietf:params:qlog:http#h4:new-event-Y 
        urn:ietf:params:qlog:http#h4:new-event-Z         

As long as both docs only define unique event types (new-event-X and -Y here), then indeed, no problem. But there's nothing stopping two independent docs from ending up with `h4:new-event-Z` independently. 

At that point, you fall back to "whoever is first to register the h4 category identifier with IANA". If that's the intent, then I can understand that and (probably) live with it. But that isn't entirely clear from the current text imo (at least the part that it's allowed to extend an existing schema. So I'd add an example to say that another new document could register `urn:ietf:params:qlog:john#cleese` event though `urn:ietf:params:qlog:john` already existed. 

I mainly want to confirm the intent here and make sure we make this as clear as possible, since even I am having a bit of a hard time fully nailing it down :)

draft-ietf-quic-qlog-main-schema.md
LPardue: That's specification via example, which is bad. The language needs to be clear enough on its own other wise people only code to the example.

draft-ietf-quic-qlog-main-schema.md
rmarx: I think the new text sufficiently covers this indeed.

draft-ietf-quic-qlog-h3-events.md
rmarx: I gave my perspective on this in a separate comment below https://github.com/quicwg/qlog/pull/415#issuecomment-2191940230

draft-ietf-quic-qlog-main-schema.md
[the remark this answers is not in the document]
rmarx: See new comment on this below :) you changed this to "event definitions" which is better, but still a bit vague.

At least the example currently makes clear that a single document can contain multiple schema, and that event types should belong to categories, which in turn should belong to a schema. 

In short: I still think this need clarification.

draft-ietf-quic-qlog-main-schema.md
LPardue: This section is all about how to extend qlog with more concrete event types. The URI stuff is ancilliary to that.

We need to start with clear requirements on what extensions can and can't do. 

Events have a name that conveys the category and type. When event names are not unique, ambiguitiy can cause tooling to fail. The language here is normative to avoid that class of problem. This problem exists even if we weren't trying to invent a way to indicate what an endpoint was generating e.g. - if we both invent a foobar:frame_sent event type with wildly different data formats.

draft-ietf-quic-qlog-main-schema.md
LPardue: agree, are you happier with something like 

```suggestion
Each category identifier in a schema is named by a URI. That URI MUST be absolute and MUST contain a fragment identifier (characters after a "#" in the URI) that indicates the category identifier. 
```

draft-ietf-quic-qlog-main-schema.md
LPardue: I don't think its confusing. For the HTTP/3 qlog I'd generate for quiche, I'm already expecting to do the following based on identifiers in the QUIC and H3 extension schema

```
~~~
"additional_event_schemas": [
                            "urn:ietf:params:qlog:quic#quic",
                            "urn:ietf:params:qlog:quic#security",
                            "urn:ietf:params:qlog:quic#recovery",
                            "urn:ietf:params:qlog:h3#h3"
                          ]
~~~

**Comment by LPardue** [comment]
Putting aside the actual text I wrote for a moment, the logical structure I have in mind is

```
schema namespace
  category identifier
    event type
```
 
Then, the `additional_event_schemas` lists URIs that must include the schema namespace and category identifier but not the event type.
 
IETF documents intended to be RFC, should use a URI of format `urn:ietf:params:qlog:<schema namespace>#<category identifier`. 

Any other legal URI is also allowed for private extensions. There is still a requirement that the URI includes both the `<schema namespace>` `#<category identifier>`. 

For IETF documents, the need to register schema namespace and category identifier (the URI is a byproduct of this, not the definition IMO, and that is what is causing some friction IMO). If a namespace already exists, it should IMO be allowed to add categories to it. IANA can deal with that. 

If somehow the IETF has adopted work items are are, in parallel, defining conflicting things (schema namespaces, category identiers, or event types) that is something I believe can be resolved through the regular standards process. Its likely a simple change that benefits both parties. I expect this to happen well in advance of the formal IANA action (i.e. during WG review, or LC).

One way to present this a bit more clearly is an IANA table that focuses more on the components and less on the URI e.g.

| Schema Namespace | Category ID | Description | Link |
|------------------|-----------------|---------------------|-----------------|-------|
| quic | recover | For events related to QUIC recovery | RFC link |
| rick | pickle |  For events related to pickled Richards | RFC link |

In other words, it won't matter if a schema namespace exists or not for practical purposes. Experts will still need to check if the addition of a category to an existing namespace makes sense for logical purposes though.

File: draft-ietf-quic-qlog-main-schema.md (hunk 26 above)
Comment (rmarx) [reply draft-ietf-quic-qlog-main-schema.md:1169]: I agree that we have to have clear requirements. I'm mainly saying that this specific text either needs to move (original suggestion) or needs to be explained better in-place (which seems your suggestion) to really make sense for the reader.

I'm guessing this will be easily solved if we clearly explain this up front:

schema namespace > category identifier > event type 

and then it's easier to use that terminology to explain this restriction.

File: draft-ietf-quic-qlog-main-schema.md (hunk 29 above)
Comment (rmarx) [reply draft-ietf-quic-qlog-main-schema.md:1180]: yes, that's much more accurate :)

File: draft-ietf-quic-qlog-main-schema.md (hunk 31 above)
Comment (rmarx) [reply draft-ietf-quic-qlog-main-schema.md:1197]: I agree that wouldn't be ideal, but then indeed we need to have better text imo :) 

Maybe something more generic up top, like:

``` urn:ietf:params:qlog:<SCHEMA-NAMESPACE>#<CATEGORY-IDENTIFIER> 

...

For example, this document defines the "main" schema namespace with two category identifiers: "generic" and "simulation".

``` 

to help explain how the concepts map to the URIs.

rmarx:
Reading your last reply, I think I correctly understand your intent, and I think it's the right way forward indeed :) So now we just need to clarify the text to make the intent clearer to new readers. As I said up-top somewhere as well, explaining things like you do above with `urn:ietf:params:qlog:<schema namespace>#<category identifier>` would already help a lot.

Another thing I think we should do, is have the main schema also define its own schema namespace ("main") with two categories ("generic" and "simulation"). Currently, these are kind of just "there" and implied... We can then use those as examples without having to go out-document, and it's more consistent/cleaner. 

It also makes it possible to indicate in `additional_event_schemas` whether or not you're actually using "generic" and "simulation" events (probably not). That DOES mean we should re-bikeshed the `additional` part of the name (and the idea of only having "extension" schemas) maybe :)

LPardue:
Yeah this all makes sense to me, #rollsupsleaves

LPardue:
closing in favor of #424, which took the same premise but ironed out the issues raised here.

Issue #261, opened 2022-11-01
Approach for CDDL extension points


rmarx:
It is expected that the initial 3 qlog documents will be extended in the future with new events and frames (e.g., for QUIC and H3 extensions like DATAGRAM or Multipath, but also for other protocols like H2, TLS and TCP).

The proposed approach is to define these qlog extensions in their own documents (either standalone or integrated into the docs that define the protocol extensions themselves). Ideally, these new documents wouldn't just extend qlog in prose only, but also properly extend the existing CDDL event and data type definitions. 

To this end, the current documents define a few [CDDL plugs/sockets/extension points](https://www.rfc-editor.org/rfc/rfc8610.html#section-3.9) to make this possible:
- [`$ProtocolEventBody`](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-main-schema-04.html#name-data)
- [`$QuicFrame`](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-quic-events-03.html#name-quic-frames)
- [`$HTTPFrame`](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-h3-events-03.html#name-httpframe)

However, when trying to define [a simple extension for the QUIC and H3 DATAGRAM frames](https://github.com/rmarx/draft-marx-quic-qlog-datagram), I found we might need more extension points, specifically for QUIC Transport Parameters and HTTP SETTINGS (the DATAGRAM RFCs use those to negotiate support). These currently don't exist and the extension document uses some not-really-compatible CDDL at this point. 

While it's not a big problem to define new extension points for those two things, it begs the question if that would be enough. Do we need even more extension points down the line (e.g., for the `ConnectionState` in [connection_state_updated](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-quic-events-03.html#name-connection_state_updated)? For `StreamState` in [stream_state_updated](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-quic-events-03.html#name-stream_state_updated)? For data handling points in [data_moved](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-quic-events-03.html#name-data_moved)? For metrics in [metrics_updated](https://www.ietf.org/archive/id/draft-ietf-quic-qlog-quic-events-03.html#name-metrics_updated)? etc.)

[...]

Issue #297, opened 2023-03-25
Logging for paced packet sending


LPardue:
Today qlog has packet_sent events and datagrams_sent events, which can be used to log the event and the time it occurred.

In some scenarios, the application generating packets or even datagrams is not in direct control of the emission of those packets to the wire. For example, a QUIC library may be called to generate a batch of packets that are intended to be sent out on a future time schedule. If the pacing is offloaded to some other component, such as the kernel, then reporting events in a qlog might be practically hard.

Something we've done in quiche is to add an additional `send_at_time` field in the packet_sent event. This contains a timestamp that the QUIC layer decided that the packet should be sent out at. We can keep that as custom extension but I was curious if others had any opinions.

LPardue:
We can close this when #400 lands

rmarx:
Small correction: not #400 but its more general spinoff, which will be tracked in #379 :)

rmarx:
Solved by #417

PR #417, opened 2024-03-18
Add extensibility for all events and official extension points

### LINKED ISSUE #379: Should we use "* text => uint64" anywhere that a type can contain extension fields?
In H3 [parameters_set](https://quicwg.org/qlog/draft-ietf-quic-qlog-h3-events.html#name-parameters_set) , we have a

```
; additional settings for grease and extensions
    * text => uint64
```

if I grok CDDL correctly, this allows us to have any number of pairs that would serialize like 

```
"my_sekret_extension": 123,
"9999": 0
```

It got me thinking a bit. This isn't explained super clearly and people might overlook it. Maybe there's more use for this in some other types. Leading to, maybe we should define this as a generic tuple type in the base spec along with some text, then we can reuse it across various events?

rmarx:
Fixes #379.
Also closes https://github.com/quicwg/qlog/issues/261, https://github.com/quicwg/qlog/issues/176, https://github.com/quicwg/qlog/issues/170, https://github.com/quicwg/qlog/issues/124, https://github.com/quicwg/qlog/issues/192, https://github.com/quicwg/qlog/issues/297.

Adds extensibility for all events through the `* $$category-name-ext` pattern first used in #400, now extended to ALL events as discussed in #379.

Makes all official protocol extension points also extensible in qlog. This was mostly already the case except for a few holdouts ($H3StreamType, $ProtocolType, $TransportError). The full list currently is:
- frames: $QuicFrame and $H3Frame and $H3Datagram
- TPs: $$quic-parametersset-extension and $$quic-parametersrestored-extension 
- settings: $$h3-parameters-extension (and inherently H3Setting by the way it's defined)
- H3 stream types: $H3StreamType
- error codes: $ApplicationError, $TransportError
- ALPN (at least qlog-level APLN; the other one is inherent in ALPNInformation): $ProtocolType

I don't think we need to make QUIC's StreamType extensible (since that's also in the RFC as always just uni- or bidirectional, with no way to change that through IANA (or even wire image, due to looking at the 2 LSBs)). Thoughts on this @LPardue ?

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -146,7 +146,10 @@ in this specification.
 
 # HTTP/3 Events {#h3-ev}
 
-HTTP/3 events extend the `$ProtocolEventData` extension point defined in {{QLOG-MAIN}}.
+HTTP/3 events extend the `$ProtocolEventData` extension point defined in
+{{QLOG-MAIN}}. Additionally, they allow for direct extensibility by their use of
+per-event extension points via the `$$` CDDL "group socket" syntax, as also
+described in {{QLOG-MAIN}}.
 
 ~~~ cddl
 H3EventData = H3ParametersSet /
@@ -436,8 +439,9 @@ Owner = "local" /
 
 ## H3Frame
 
-The generic `$H3Frame` is defined here as a CDDL extension point (a "socket"
-or "plug"). It can be extended to support additional HTTP/3 frame types.
+The generic `$H3Frame` is defined here as a CDDL extension point (a "type
+socket" or "type plug"). It can be extended to support additional HTTP/3 frame
+types.
 
 ~~~ cddl
 ; The H3Frame is any key-value map (e.g., JSON object)
@@ -466,9 +470,10 @@ $H3Frame /= H3BaseFrames
 
 ## H3Datagram
 
-The generic `$H3Datagram` is defined here as a CDDL extension point (a "socket"
-or "plug"). It can be extended to support additional HTTP/3 datagram types. This
-document intentionally does not define any specific HTTP/3 Datagram types.
+The generic `$H3Datagram` is defined here as a CDDL extension point (a "type
+socket" or "type plug"). It can be extended to support additional HTTP/3
+datagram types. This document intentionally does not define any specific HTTP/3
+Datagram types.
 
 ~~~ cddl
 ; The H3Datagram is any key-value map (e.g., JSON object)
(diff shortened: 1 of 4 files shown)


rmarx:
TODO for myself: we also need to make PacketHeader extensible (e.g., to be able to add lossbits, new fields for QUIC v3 etc.)

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -214,13 +224,17 @@ H3Parameters = {
 
     ; additional settings for grease and extensions
     * text => uint64
+
+    * $$h3-parameters-extension
```
LPardue, line 228: can we remove the `text => uint` now that we have $$h3-parameters-extension?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -651,69 +650,150 @@ JSON serialization example:
 
 ## Data {#data-field}
 
-An event's "data" field is a generic object. It contains the per-event metadata and its
-form and semantics are defined per specific sort of event. For example, data
-field value definitions for QUIC and HTTP/3 can be found in {{QLOG-QUIC}} and
-{{QLOG-H3}}.
+An event's "data" field is a generic key-value map (e.g., JSON object). It
```
LPardue, line 653: Is there any danger that mentioning JSON object here could confusing people about what serialization formats are allowed?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -651,69 +650,150 @@ JSON serialization example:
...
+; if document A, B and C are then used in conjunction,
+; the combined MyCategoryEventX CDDL is equivalent to this:
+MyCategoryEventX = {
+    field_a: uint8
+
+    ? additional_field_b: bool
+    ? additional_field_c: text
+}
+~~~~~~~~
+{: #groupsocket-extension-example title="Example of using a generic CDDL group socket to extend an existing event data definition"}
```
LPardue, line 696: This example makes me realize that qlog event data field order cannot be significant. I don't think we state that anywhere already?

draft-ietf-quic-qlog-main-schema.md
rmarx: I thought about it, but I dont' think so... we discuss this in detail in the serialization section and JSON is the main provided option there, so I think it's fine to have it as an example here.

draft-ietf-quic-qlog-main-schema.md
rmarx: Good point. Tracking that here now: https://github.com/quicwg/qlog/pull/418

draft-ietf-quic-qlog-main-schema.md
LPardue: agree

rmarx:
Did some final work on this today. Should be ready for final review + merge @LPardue @marten-seemann

LPardue (approved):

PR #419, opened 2024-06-24
Make intent of triggers for connection_closed clearer

### LINKED ISSUE #410: Connectivity::Connection_closed application trigger type is confusing
```
ConnectivityConnectionClosed = {

    ; which side closed the connection
    ? owner: Owner
    ? connection_code: TransportError /
                       CryptoError /
                       uint32
    ? application_code: $ApplicationError /
                        uint32
    ? internal_code: uint32
    ? reason: text
    ? trigger:
        "clean" /
        "handshake_timeout" /
        "idle_timeout" /
        ; this is called the "immediate close" in the QUIC RFC
        "error" /
        "stateless_reset" /
        "version_mismatch" /
        ; for example HTTP/3's GOAWAY frame
        "application"
}
```

GOAWAY doesn't close the connection, so I find this example a bit confusing.

In quiche, I would be able to close a connection using transport or an application CONNECTION_CLOSE and log that information in the other fields like `connection_code` etc. 

Maybe we should remove that trigger variant?

rmarx:
Closes #410 

@LPardue PTAL. I still think having an `application` value here is useful (e.g., user force-closed the connection somehow, or application decided to abandon it without waiting for clean shutoff without additional info). But as discussed on #410 the GOAWAY example was probably incorrect.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -347,7 +347,7 @@ ConnectivityConnectionClosed = {
 ~~~
 {: #connectivity-connectionclosed-def title="ConnectivityConnectionClosed definition"}
 
-Loggers should use the most descriptive trigger for a `connection_closed` event
+Loggers SHOULD use the most descriptive trigger for a `connection_closed` event
 that they are able to deduce. This is often clear at the peer closing the
 connection (and sending the CONNECTION_CLOSE), but can sometimes be more opaque
 at the receiving end.
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -330,7 +330,6 @@ ConnectivityConnectionClosed = {
         "error" /
         "stateless_reset" /
         "version_mismatch" /
-        ; for example HTTP/3's GOAWAY frame
         "application"
 
     * $$connectivity-connectionclosed-extension
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -304,6 +304,16 @@ deployments, it can be useful to have a single event representing a
 provide more information. Furthermore, it is useful to log closures due to
 timeouts, which are difficult to reflect using the other options.
 
+The `connection_closed` event is intended to be logged either when the local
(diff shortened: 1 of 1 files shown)


rmarx:
Discussed during editors' meeting. Discussion on how useful `clean` is. Discussion on when connection_close should be emitted (before draining or after draining, if there's a draining involved). Marten feels it should be when draining starts, Lucas has no opinion.

Robin will go back and see what would be most useful and what makes sense with errors logging etc. (i.e., should be emit ConnectionClosed twice or use ConnectionStateUpdated for some things)

LPardue (approved):

rmarx:
Ok, after spending WAY too much time re-reading the connection close stuff in RFC9000 and 9114, I think I now have something that's still a bit vague, but at least correct :)

The main problem apparently (and why "clean" was needed), is because you have situations where you as a peer receive a CONNECTION_CLOSE where you're not 100% sure what triggered it... (e.g., a transport CONNECTION_CLOSE (0x1c) with transport error code NO_ERROR) is probably clean, but you're not quite sure (could be it was an application-level error on the other end that they simply didn't pipe through correctly).  

So I've now renamed that to "unspecified" and added text to say you should try to be as specific as possible... hope that is enough :) 

PTAL @LPardue

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -323,21 +333,25 @@ ConnectivityConnectionClosed = {
...
+        "stateless_reset" /
+        ; when it is unclear what triggered the CONNECTION_CLOSE
+        "unspecified"
 
     * $$connectivity-connectionclosed-extension
 }
 ~~~
 {: #connectivity-connectionclosed-def title="ConnectivityConnectionClosed definition"}
 
+Loggers should use the most descriptive trigger for a `connection_closed` event
```
LPardue, line 350: Should or SHOULD

rmarx: Discussed in editors' meeting. Change to `SHOULD`

LPardue:
```suggestion
Loggers SHOULD use the most descriptive trigger for a `connection_closed` event
```

PR #424, opened 2024-06-29
Behold - schema

LPardue:
This looks like hell to review but a lot of it is moving text around, so a visual diff is probably the most straightforward.

The high-level (pun-intended) goal of this PR is to realize the `main-schema` document is not merely a schema. Its the document that defines the core concepts behind its structured logging, methodology, requirements for loggers and tools, AND a few schema. 
Moving the document away from referring to its entire self as a schema, makes it easier to isolate the parts we mostly expect people to extend - the event definitions.

I found it incredibly hard to talk about this with the old terminology of generic event class and specific data. Especially when combined with the concept of "event definition documents". That's been reflected into prior attempts at wrangling the extensibility of events. So I've changed that to abstract Event class and concrete event types. I suspect those won't be that popular either but they work for this old OOP survivor. 

The PR isn't 100% complete but it does implement the schema > category > event structure that we seemed to align on, and ports over a lot more from https://github.com/quicwg/qlog/pull/415. If we think this is a suitable direction, putting the finishing touches like IANA policy and updating QUIC/H3 docs isn't that much more work.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -232,17 +232,17 @@ use case. For example, a QUIC packet being sent or received. This document
 declares an abstract Event class ({{abstract-event}}) containing common fields, which
 all concrete events derive from. Concrete events are defined by event schema
 that declare a namespace, consisting of one or more categories, containing one
-or more related event types. For example, this document defines the `std` event
+or more related event types. For example, this document defines the generic event
 schema structured as:
 
-* `std` namespace
-  * `generic` category
+* `gen` namespace
+  * `loglevel` category
     * `error` event type
     * `warning` event type
     * `info` event type
     * `debug` event type
     * `verbose` event type
-  * `simulation` category
+  * `sim` category
     * `scenario` event type
     * `marker` event type
 
@@ -1014,15 +1014,15 @@ a listed category identifier. Tools MUST NOT treat this as an error; see
 {{tooling}}.
 
 In the following hypothetical example, a qlog file contains events belonging to
-the standard event schema ({{std-event-schema}}), an event schema named `rick`
+the standard event schema ({{generic-event-schema}}), an event schema named `rick`
 specified in a hypothetical RFC, and a private event schema named pickle. The
 standardized categories use a URN format, the private categories use a URI with
 domain name.
 
 ~~~
 "event_schemas": [
-  "urn:ietf:params:qlog:events:std#generic,
-  "urn:ietf:params:qlog:events:std#simulation,
+  "urn:ietf:params:qlog:events:gen#loglevel,
+  "urn:ietf:params:qlog:events:gen#sim,
   "urn:ietf:params:qlog:events:rick#roll",
   "urn:ietf:params:qlog:events:rick#astley",
   "urn:ietf:params:qlog:events:rick#moranis",
@@ -1039,8 +1039,8 @@ Event schema defined by RFCs MUST register all categories in the "qlog event
 category URIs" registry and SHOULD use a URN of the form
 `urn:ietf:params:qlog:events:<namespace identifier>#<category identifier>`,
 where `<category identifier>` is globally unique. For example, this document
-defines the standard event schema ({{std-event-schema}}) that uses the `std`
-namespace containing the `generic` and `simulation` categories. Other examples
+defines the standard event schema ({{generic-event-schema}}) that uses the `gen`
+namespace containing the `loglevel` and `sim` categories. Other examples
 of event schema define the `quic` {{QLOG-QUIC}} and `h3` {{QLOG-H3}} namespaces.
 
 Private or non-standard event categories can use other URI formats. URIs that
@@ -1300,29 +1300,29 @@ additional events is typically avoided. Exceptions have been made for common
 events that benefit from being easily identifiable or individually logged (for
 example `packets_acked`).
 
-# The Standard Event Schema {#std-event-schema}
+# The Generic Event Schema {#generic-event-schema}
 
-The standard event schema defines categories and event types that are common across
-protocols, applications, and use cases. The schema namespace is "std".
+The generic event schema defines categories and event types that are common across
+protocols, applications, and use cases. The schema namespace is "gen".
 
-## Generic Events
+## Log Level Events {#loglevel-events}
 
 In typical logging setups, users utilize a discrete number of well-defined logging
-categories, levels or severities to log freeform (string) data. This generic
-events category replicates this approach to allow implementations to fully replace
+categories, levels or severities to log freeform (string) data. The log level
+event category replicates this approach to allow implementations to fully replace
 their existing text-based logging by qlog. This is done by providing events to log
 generic strings for the typical well-known logging levels (error, warning, info,
-debug, verbose). The category identifier is "generic".
+debug, verbose). The category identifier is "loglevel".
 
 ~~~ cddl
-GenericEventData = GenericError /
-                GenericWarning /
-                GenericInfo /
-                GenericDebug
+LogLevelEventData = LogLevelError /
+                    LogLevelWarning /
+                    LogLevelInfo /
+                    LogLevelDebug
 
-$ProtocolEventData /= GenericEventData
+$ProtocolEventData /= LogLevelEventData
 ~~~
-{: #generic-events-def title="GenericEventData and ProtocolEventData extension"}
+{: #loglevel-events-def title="LogLevelEventData and ProtocolEventData extension"}
 
 The event types are further defined below, their identifier is the heading name.
 
@@ -1332,14 +1332,14 @@ Used to log details of an internal error that might not get reflected on the
 wire. It has Core importance level.
 
 ~~~ cddl
-GenericError = {
+LogLevelError = {
     ? code: uint64
     ? message: text
 
-    * $$generic-error-extension
+    * $$loglevel-error-extension
 }
 ~~~
-{: #generic-error-def title="GenericError definition"}
+{: #loglevel-error-def title="LogLevelError definition"}
 
 ### warning
 
@@ -1347,14 +1347,14 @@ Used to log details of an internal warning that might not get reflected on the
 wire. It has Base importance level; see {{importance}}.
 
 ~~~ cddl
-GenericWarning = {
+LogLevelWarning = {
     ? code: uint64
     ? message: text
 
-    * $$generic-warning-extension
+    * $$loglevel-warning-extension
 }
 ~~~
-{: #generic-warning-def title="GenericWarning definition"}
+{: #loglevel-warning-def title="LogLevelWarning definition"}
 
 ### info
 
@@ -1363,13 +1363,13 @@ logging format but still want to support unstructured string messages. The event
 has Extra importance level; see {{importance}}.
 
 ~~~ cddl
-GenericInfo = {
+LogLevelInfo = {
     message: text
 
-    * $$generic-info-extension
+    * $$loglevel-info-extension
 }
 ~~~
-{: #generic-info-def title="GenericInfo definition"}
+{: #loglevel-info-def title="LogLevelInfo definition"}
 
 ### debug
 
@@ -1378,13 +1378,13 @@ logging format but still want to support unstructured string messages. The event
 has Extra importance level; see {{importance}}.
 
 ~~~ cddl
-GenericDebug = {
+LogLevelDebug = {
     message: text
 
-    * $$generic-debug-extension
+    * $$loglevel-debug-extension
 }
 ~~~
-{: #generic-debug-def title="GenericDebug definition"}
+{: #loglevel-debug-def title="LogLevelDebug definition"}
 
 ### verbose
 
@@ -1393,13 +1393,13 @@ logging format but still want to support unstructured string messages. The event
 has Extra importance level; see {{importance}}.
 
 ~~~ cddl
-GenericVerbose = {
+LogLevelVerbose = {
     message: text
 
-    * $$generic-verbose-extension
+    * $$loglevel-verbose-extension
 }
 ~~~
-{: #generic-verbose-def title="GenericVerbose definition"}
+{: #loglevel-verbose-def title="LogLevelVerbose definition"}
 
 ## Simulation Events
 
@@ -1408,7 +1408,7 @@ interoperability or benchmarking tests, in which the test situations can change
 over time. For example, the network bandwidth or latency can vary during the test,
 or the network can be fully disable for a short time. In these setups, it is
 useful to know when exactly these conditions are triggered, to allow for proper
-correlation with other events. The category identifier is "simulation".
+correlation with other events. The category identifier is "sim".
 
 ~~~ cddl
 SimulationEventData = SimulationScenario /
@@ -1915,7 +1915,8 @@ event categories. It has the following format:
 
 | Event Category URI | Description | Reference |
 |||||
-| urn:ietf:params:qlog:events:std#generic | Well-known logging levels for free-form text. | {{generic-events}} |
+| urn:ietf:params:qlog:events:gen#loglevel | Well-known logging levels for free-form text. | {{loglevel-events}} |
+| urn:ietf:params:qlog:events:gen#sim | Events for simulation testing. | {{loglevel-events}} |
 
 
 --- back
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -261,8 +261,7 @@ LogFile = {
     ? description: text
     event_schemas: [+text]
 
-    ; can contain any amount of custom fields
-    * text => any
+    * $$logfile-extensions
 }
 ~~~
 {: #abstract-logfile-def title="LogFile definition"}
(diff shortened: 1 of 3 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1062,98 +923,338 @@ trace has a different value for a given field, this field MUST NOT be added to
...
-still indicate the non-truncated lengths when used for debugging purposes.
+Event schema MUST define a registered non-empty namespace of type `text`.
 
-This document does not specify explicit header_length or trailer_length fields.
-In protocols without trailers, header_length can be calculated by subtracting
-the payload_length from the length. In protocols with trailers (e.g., QUIC's
-AEAD tag), event definition documents SHOULD define how to support header_length
-calculation.
+Event categories MUST belong to a single event schema. The MUST have a
+registered non-empty globally-unique identifier of type `text` and MUST have a
```
rmarx, line 938: This is not entirely how I understood things... categories do not have to be GLOBALLY unique, just within their specific namespace (e.g., you could have  namespaceX:categoryA and also namespaceY:categoryA). As such, the COMBINATION of namespace + category identifier should be globally unique. 

Additionally: I think it's better to say `Event categories MUST belong to a single event namespace` instead of to a single event `schema`... helps to reinforce the URI structure later.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1062,98 +923,338 @@ trace has a different value for a given field, this field MUST NOT be added to
...
+single dereferencable URI. That URI MUST be absolute and MUST indicate the
+category identifier using a fragment identifier (characters after a "#" in the
+URI). For event categories in schema defined by RFCs, the URI SHOULD be a URN of
+the form `urn:ietf:params:qlog:<schema namespace>#<category identifier>`.
```
rmarx, line 942:
```suggestion
the form `urn:ietf:params:qlog:<namespace identifier>#<category identifier>`.
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1062,98 +923,338 @@ trace has a different value for a given field, this field MUST NOT be added to
...
-    ? data: hexstring
-}
-~~~
-{: #raw-info-def title="RawInfo definition"}
+Concrete events MAY extend any part of the abstract Event class, including extending the "data" field ({{data-field}}) of adding custom fields.
 
-The RawInfo:data field can be truncated for privacy or security purposes, see
-{{truncated-values}}. In this case, the length and payload_length fields should
-still indicate the non-truncated lengths when used for debugging purposes.
+Event schema MUST define a registered non-empty namespace of type `text`.
```
rmarx, line 935:
```suggestion
Event schema MUST define a registered non-empty namespace identifier of type `text`.
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1062,98 +923,338 @@ trace has a different value for a given field, this field MUST NOT be added to
...
+problems when domain names change ownership.)
 
-# Common events and data classes
+The registration requirements for extension schema URIs are detailed in
+{{iana}}.
 
-There are some event types and data classes that are common across protocols,
-applications, and use cases. This section specifies such common definitions.
+Concrete event types MUST belong to a single event category and MUST have a
+non-empty identifier of type `text`.
```
rmarx, line 953:
```suggestion
non-empty name of type `text`.
```

since you later use "event type identifier" to indicate just the "type name" (second part of the `name`).

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1062,98 +923,338 @@ trace has a different value for a given field, this field MUST NOT be added to
...
+                      "urn:ietf:params:qlog:rick#moranis",
+                      "urn:ietf:params:qlog:john#doe",
+                      "urn:ietf:params:qlog:john#candy",
+                      "urn:ietf:params:qlog:john#goodman",
+                      "https://example.com/032024/pickle.html#pepper",
+                      "https://example.com/032024/pickle.html#lilly",
+                      "https://example.com/032024/pickle.html#rick",
+                    ]
+~~~
+{: #event-categories title="Example event_categories serialization"}
```
rmarx, line 978: At no point has it been made clear yet how loggers MUST/SHOULD fill the `event_categories` field, nor how tools should handle "missing" URIs (i.e., unknown event types). I think this should be explicitly called out.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -209,6 +210,39 @@ The main tenets for the qlog schema design are:
   group_ids can be used to tag events to a particular context)
 * Metadata is stored together with event data
 
+This is achieved by a logical logging hierarchy of:
+
+* Log file
+  * Trace(s)
+    * Event(s)
+
+Two log file schema are defined: QLogFile ({{qlog-file-schema}}) and QLogFileSeq
```
rmarx, line 219: I don't particularly like the "file schema" vs "event schema" as we currently have it... I understand WHY you're doing it that way, but it seems a bit incongruous/inconsistent (e.g., we don't explain how/why to define new FILE schema, nor have we ways to extend existing FILE schema, in contrast to EVENT schema). 

maybe this is just me coming from the "old" way of using "schema", but it doesn't feel completely right semantically to me atm.

rmarx (changes requested):
Thanks @LPardue for this effort :) 

I've tried to stick mostly to the main points for my first pass. In general, I like the new approach, and you've bypassed a lot of the minor issues from the previous version in #415. 

In general, I think my comments here are not blockers or major issues, just things that could do with further refinement / clarification. Let's discuss these tomorrow in the editors' meeting and then I can do a final "editorial" pass to fix some consistency/spelling issues later :)

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -61,12 +62,14 @@ informative:
...
-logging format. It allows easy sharing of data, benefitting common debug and
-analysis methods and tooling. The high-level schema is independent of protocol;
-separate documents extend qlog for protocol-specific data. The schema is also
-independent of serialization format, allowing logs to be represented in many
-ways such as JSON, CSV, or protobuf.
+qlog provides extensible structured logging for network protocols, allowing for
+easy sharing of data that benefits common debug and analysis methods and
+tooling. This document describes key concepts of qlog: formats, files, traces,
+events, and extension points. In this definition are the high-level log file
+schemas, and main event schema. Requirements and guidelines for creating
```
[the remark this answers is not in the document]
LPardue: Yeah it felt to me more lime this is stdlib than `main` but I refrained from changing it since the pr was some huge anyway. I'm open to a different value

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -97,32 +100,35 @@ interoperability failure between distinct implementations. A lack of a common
...
-provides a shareable, aggregatable and structured logging format. This
-high-level schema is independent of protocol, with logging entries for specific
-protocols and use cases being defined in other documents (see for example
-{{QLOG-QUIC}} for QUIC and {{QLOG-H3}} for HTTP/3-related event definitions).
+qlog is an extensible structured logging for network protocols that allows for
+easy sharing of data that benefits common debug and analysis methods and
+tooling. This document describes key concepts of qlog: formats, files, traces,
+events, and extension points. In this definition are the high-level log file
+schemas, and main event schema. Requirements and guidelines for creating
+protocol-specific event schema are also presented. Accompanying documents define event schema for QUIC ({{QLOG-QUIC}}) and HTTP/3 ({{QLOG-H3}}).
```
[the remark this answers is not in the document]
LPardue: This is mostly me having a blind spot for using the correct term no matter how hard I try :D

draft-ietf-quic-qlog-main-schema.md
LPardue: That creates a whole class of problem of ambiguity I'd like to avoid. If events dont self identify the namespace, and we dont make declaration of all schema a MUST, its a recipe for clashes where people add a super generic category like `util` and have an event name like `util:do_common_thing`. 

Instead if we operate the old school C methodology of embedding the unique token in the signature, we don't need to rely on namespace mangling. 

If we don't want to go this way, that's OK but we need a way to avoid name collisions when the event names could clash

draft-ietf-quic-qlog-main-schema.md
LPardue: Very good poont

draft-ietf-quic-qlog-main-schema.md
rmarx: Discussed during editors' meeting.

Lucas is in the right here, should be ok as the current text states. Robin was conflating with something else.

Robin would like us to explicitly call out how things can be extended though:
- CANNOT add new events to existing categories in other documents
- CANNOT create new categories that have the same name as existing categories in other documents (this is basically what the current text above says)
- CAN create new categories in existing NAMESPACES as long as the category names are globally unique

And these should also be supported by a concrete example.

draft-ietf-quic-qlog-main-schema.md
rmarx: Discussed in editors' meeting. 

Make explicit in new `file_schema` field with its own URI namespace and IANA registration. Allows us to also get rid of `qlog_version` (implicit in file_schema now)

Rename `event_categories` to `event_schemas` maybe for consistency with `file_shema`

```
urn:ietf:params:qlog:file:
urn:ietf:params:qlog:events:

file_schema: "urn:ietf:params:qlog:file:sequential",
event_schemas: [
    "urn:ietf:params:qlog:events:rick#roll",
    "urn:ietf:params:qlog:events:rick#astley",
    "urn:ietf:params:qlog:events:rick#moranis",
    "urn:ietf:params:qlog:events:john#doe",
    "urn:ietf:params:qlog:events:john#candy",
    "urn:ietf:params:qlog:events:john#goodman",
    "https://example.com/032024/pickle.html#pepper",
    "https://example.com/032024/pickle.html#lilly",
    "https://example.com/032024/pickle.html#rick",
]
```

draft-ietf-quic-qlog-main-schema.md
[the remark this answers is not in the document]
rmarx: Discussed during editors' meeting. Should be `std` as namespace identifier.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -244,6 +246,57 @@ schema structured as:
...
+defined. Each concrete log file format derives from this, extending it by
+defining semantics and any custom fields.
+
+~~~ cddl
+LogFile = {
+    file_schema: text
+    qlog_serialization_format: text
+    ? title: text
+    ? description: text
+    event_schema: [+text]
```
rmarx, line 262: should this be `event_schemas` instead (making it implicitly clear there is 1 `file_schema` but multiple `event_schemaS`?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -244,6 +246,57 @@ schema structured as:
...
+
+    ; can contain any amount of custom fields
+    * text => any
```
rmarx, line 265: You probably know this, but this should be something like `$$logfile-extensions` with CDDL

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -244,6 +246,57 @@ schema structured as:
...
+
+A Log file is intended to contain a collection of events that are in some way
+related. An abstract LogFile class containing fields common to all log files is
+defined. Each concrete log file format derives from this, extending it by
+defining semantics and any custom fields.
+
+~~~ cddl
+LogFile = {
+    file_schema: text
+    qlog_serialization_format: text
```
rmarx, line 259: I think we can probably just leave this as `serialization_format`, since we don't have the `qlog_` prefix anywhere else anymore :)

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -244,6 +246,57 @@ schema structured as:
...
+and their values SHOULD be in the first 256 characters/bytes of the resulting
+log file.
+
+The required "qlog_serialization_format" field indicates the serialization
+format using a media type {{!RFC2046}}. It is case-insensitive.
+
+The optional "title" and "description" fields provide additional free-text
+information about the file.
+
+The required "event_schema" field is described in {{event-types-and-schema}}.
```
rmarx, line 283: Give a short description here as well about the goal of this field (just so readers get a high-level overview from this section)

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -244,6 +246,57 @@ schema structured as:
...
+
+## Concrete Schema URI {#schema-uri}
+
+Concrete log file schemas MUST identify themselves using a URI.
+
+Log schemas defined by RFCs SHOULD be URN of the form
+`urn:ietf:params:qlog:file:<schema-identifier>`, where `<schema-identifier>` is
+a globally-unique name. URN MUST be registered with IANA; see {{iana}}.
+
+Private or non-standard event categories can use other URI formats. URIs that
```
rmarx, line 293: copy-paste whoopsie :) should be `or non-standard log file schemas`

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -244,6 +246,57 @@ schema structured as:
...
+
+The required "event_schema" field is described in {{event-types-and-schema}}.
+
+## Concrete Schema URI {#schema-uri}
+
+Concrete log file schemas MUST identify themselves using a URI.
+
+Log schemas defined by RFCs SHOULD be URN of the form
+`urn:ietf:params:qlog:file:<schema-identifier>`, where `<schema-identifier>` is
+a globally-unique name. URN MUST be registered with IANA; see {{iana}}.
```
rmarx, line 291: Should we mention here again/already that this document registers 2 log file schema URNs?

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -244,6 +246,57 @@ schema structured as:
...
+The optional "title" and "description" fields provide additional free-text
+information about the file.
+
+The required "event_schema" field is described in {{event-types-and-schema}}.
+
+## Concrete Schema URI {#schema-uri}
+
+Concrete log file schemas MUST identify themselves using a URI.
+
+Log schemas defined by RFCs SHOULD be URN of the form
```
rmarx, line 289: Log file schemas

rmarx (approved):
I like the general direction of re-using media types here; feels clean.

The rest is also as discussed, minus a few nits which I've listed here, so let's continue in this direction :)

draft-ietf-quic-qlog-main-schema.md
LPardue: agreed, done

draft-ietf-quic-qlog-main-schema.md
LPardue: yep, done

draft-ietf-quic-qlog-main-schema.md
LPardue: fixed

draft-ietf-quic-qlog-main-schema.md
LPardue: fixed

draft-ietf-quic-qlog-main-schema.md
LPardue: Makes sense but then the concrete log files need to extend that right? I gave it a go, PTAL

draft-ietf-quic-qlog-main-schema.md
LPardue: done

draft-ietf-quic-qlog-main-schema.md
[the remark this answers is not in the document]
rmarx: Discussed during editors' meeting.

Change to `generic` namespace and make category name `loglevel` instead :)

LPardue:
Final actions to complete (from the editors call)

- [ ] - fix group socket CDDL weirdness
- [ ] - change `generic` to `loglevel`, change `std` to `generic`
- [ ]  - add "MAY register a URI in the "qlog eventschema" registry but MUST NOT use a URN of the form" reviewer guidance
- [ ] - restore inadvertently deleted H3 text into intro section
- [ ] - Open an issue for URI safe schema and category names etc.

draft-ietf-quic-qlog-main-schema.md
rmarx: So turns out my suggestion of using the group socket doesn't actually work :) We need to use the CDDL "unwrapping operator" `~` instead. This will "copy and paste" all the fields from the operand struct over to this new struct. This way, we CAN define new struct names for QlogFile and QlogFileSeq :) 

So the correct CDDL definitions for the 3 structs are (note: removed `$$logfile-extensions` from `LogFile`!)

```
~~~ cddl
LogFile = {
    file_schema: text
    serialization_format: text
    ? title: text
    ? description: text
    event_schemas: [+text]
}
~~~
{: #abstract-logfile-def title="LogFile definition"}
```

```
~~~ cddl
QlogFile = {
    ~LogFile
    ? traces: [+ Trace /
                TraceError]
}
~~~
{: #qlog-file-def title="QlogFile definition"}
```

```
~~~ cddl
QlogFileSeq = {
    ~LogFile
    trace: TraceSeq
}
~~~
{: #qlog-file-seq-def title="QlogFileSeq definition"}
```


Note that we used this before in H3, but I removed those for clarity. There, I also had some clarifying text (much like we have for the sockets), like this:

> As a reminder the CDDL unwrap operator (~), see {{?RFC8610}}), copies the fields
from the referenced type (H3Parameters) into the target type directly, extending the
target with the unwrapped fields.

Maybe a good idea to add something like that in here as well, to clarify what we're doing (replacing the current text around `$$logfile-extensions`)?

PR #433, opened 2024-07-30
Its about time

LPardue:
Don't look too deeply into the commits on this PR, its basically forking off https://github.com/quicwg/qlog/pull/290 and doing a bunch of changes to reflect one of the suggestions made at IETF 120. 

More concretely, the observation I made on the flight home is that _all_ of the timestamps are a duration relative to something. Where we use the term "absolute" timestamp, it might be reasonable to assume that the value is actually an absolute timestamp, not the millis elapsed since some point in the past that may or may not be unix. So this PR goes nuclear and changes the TimeFormat values to relative_to_epoch, relative_to_epoch_elapsed, and relative_to_last_event. We can bikeshed on these of course.

What I've taken from past discussion and the 120 meeting is the suggestion to declare the used epoch in the log format, together with a human friendly "wall clock time" that would be something like an iso 8601 format (but I was offline on the plane so couldn't cite it or give proper examples). 

Monotonic folks can declare a "monotonic" epoch, and then relative times to that are completely opaque thing. Its up to the loggers to decide how far off the monotonic epoch they want their times to be. Its recommended they use a wall clock time to help give an indication of when things occurred.


rmarx:
So, having read this and the previous issues (mainly #198) etc. I'm wondering if this can just be simplified considerably, as such:

```
ReferenceTime = {
    start: ISO8601Time / "unknown" .default "1970-01-01T00:00:00.000Z"
    clock_type: "system" / "monotonic" / text .default "system"
    ? wall_clock_time: text
}

; ISO8601 UTC string at millisecond resolution. i.e., YYYY-MM-DDTHH:mm:ss.sssZ
ISO8601Time = string

TimeFormat = "relative_to_start" / 
             "relative_to_previous_event" .default "relative_to_start"
```

IIUC, the main issue you're trying to solve is that things like [`rust:Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html) and monotonic clocks in general don't necessarily neatly map to "real time" and thus can't be transformed into "clear" timestamps that can be compared across traces/systems/...

Another good point in #198 is that it's not possible to define another epoch besides the Unix epoch against which to start values. 

I believe the above solves both.

---

The current text somewhat conflates what an epoch is in my opinion (i.e., "monotonic" doesn't make much sense for an epoch... it's more a `clock_type` than an epoch. A monotonic clock would have an epoch of `unknown` or pedantically `dont_care`). The original intent also wasn't to define an epoch as such, just to identify what the `absolute` and `delta` and `reference_time` timestamps were relative to, and the UNIX epoch was a simple thing to use. 

As such, I'd rather have a `start` time instead of `epoch`, which also allows us to do away with the `elapsed_time` field and simplifies the `TimeFormat`. If you would have the UNIX epoch in 1970 with an `elapsed_time` of 30 years, you'd just log the ISO8601 timestamp for 2000 in `start` (I don't see much reason not to have the qlogger put the proper ISO8601 timestamp in there while generating the file, rather than having the tools reconstruct it later from epoch + elapsed). If you then have a monotonic clock, or you just don't know the start time, or want to obfuscate it, you just say "unknown" (and potentially fill in the `wall_clock_time` text if you do want to give a hint).

--- 

I would of course still keep in the discussion on monotonic clocks and how to take care when using those. I would also have the new `clock_type` field for people who want to make explicitly clear this is a special kind of clock vs the "normal" system timer. But while I feel that's important to discuss in the prose, I don't think that should be key to how the format deals with time.

---

A few remarks on the current text (probably better done as direct comments...):

1. millisecond resolution

> It is a duration measured from some point in time, in millisecond resolution.

I'm not sure how this works with the monotonic clocks? with the [rust Instant example](https://doc.rust-lang.org/std/time/struct.Instant.html), it says 

> "instants are not guaranteed to be steady. In other words, each tick of the underlying clock might not be the same length (e.g. some seconds may be longer than others)"

so those milliseconds aren't "really" milliseconds? but they ARE ms resolution (probably)? So I don't think this sentence holds for monotonic clocks and should be nuanced?

2. Structure

Some of the prose should be grouped/ordered differently in my opinion, but we can discuss that once we have consensus on the general approach. 


Let's discuss this on Wednesday :)

LPardue:
Treating you comment in order, I've pushed a commit that

1. Takes the idea of using an explicit date/time value, but chosen RFC 3339 as an IETF profile of ISO8601.
2. Add the clock_type field as suggested
3. I've kept the term epoch, since that is a term of art for any fixed date/time for use as a reference. It also helps disambiguate it from anything else in scope of a qlog start (like the qlog start :laughing: ).
4. Reduced the timeformat to two types as suggested
5. Highlighted that the time field units are ???
  - The field type is float64, how are people actually supposed to use this? In my implementation I log the duration elapsed between two arbitrary times, using https://doc.rust-lang.org/std/time/struct.Duration.html#method.as_secs_f64, which is seconds with fractional nanoseconds. Yet a lot of the old examples used milliseconds with no fractional component. In section 1.2 the document states "All timestamps and time-related values (e.g., offsets) in qlog are logged as `float64` in the millisecond resolution." - should that just be moved here where its most relevant? I think this might address your open questions about the resolution. 

While your remark about the implementation detail of the monotoic clock is technically direct, I think its not super relevant for this PR. Depending how we answer (4), the timestamp can represent time in units to the resolution of milliseconds (or even nanos?), whether an implementation used that accurately is tangential. Windows is famous for having a timer resolution of ~16ms  - see https://randomascii.wordpress.com/2020/10/04/windows-timer-resolution-the-great-rule-change/. If you ever looked at logs from these platforms and noticed weird batching, its because the timer resolution != logging time field resolution.

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
 ## Timestamps {#time-based-fields}
 
-An event's "time" field indicates the timestamp at which the event occurred. Its value is
-typically the Unix timestamp since the 1970 epoch (number of milliseconds since
-midnight UTC, January 1, 1970, ignoring leap seconds). However, qlog supports two
-more succinct timestamps formats to allow reducing file size. The employed format
-is indicated in the "time_format" field, which allows one of three values:
-"absolute", "delta" or "relative".
+Each event MUST include a "time" field to indicate the timestamp that it
+occurred. It is a duration measured from some point in time, in units of ???.
```
rmarx, line 670:
```suggestion
occurred. It is a duration measured from some point in time; its units depend on the type of clock chosen and system used. Typically however, a millisecond resolution is employed.
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
-             "delta" /
-             "relative"
+ReferenceTime = {
+    clock_type: "system" / "monotonic" / text .default "system"
+    epoch: RFC3339DateTime / "unknown" .default "1970-01-01T00:00:00.000Z"
+
+    ? wall_clock_time: RFC3339DateTime
+}
+
+RFC3339DateTime = text
```
rmarx, line 691:
```suggestion
; RFC 3339 UTC string at millisecond resolution. i.e., YYYY-MM-DDTHH:mm:ss.sssZ
RFC3339DateTime = text
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
-The first option is good for stateless loggers, the second and third for stateful
-loggers. The third option is generally preferred, since it produces smaller files
-while being easier to reason about. An example for each option can be seen in
-{{time-format-ex}}.
+The required "clock_type" field represents the type of clock used for time
+measurements. The value "system" represents a clock that uses system time,
+commonly measured against a chosen or well-known epoch. System time can jump
+forward or back. The value "monotonic" represents a clock that uses monotonic
+time, measured against an arbitrary and opaque epoch. Monotonic time is
+generally guaranteed to never go backwards.
```
rmarx, line 700:
```suggestion
commonly measured against a chosen or well-known epoch. However, depending on the system, System time can potentially jump forward or back. In contrast, a clock using monotonic time is generally guaranteed to never go backwards. The value "monotonic" represents such a clock. 
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
+When using the "monotonic" clock type, the epoch field MUST have the value
+"unknown".
 
-The relative approach will:
-- set the reference_time to 1500 in "common_fields"
-- use: 0, 5, 22, 88
-~~~~~~~~
-{: #time-format-ex title="Three different approaches for logging timestamps"}
+The optional "wall_clock_time" field can be used to provide an approximate
+date/time value that logging commenced if the epoch value is "unknown". It uses
```
rmarx, line 710:
```suggestion
date/time value that logging commenced at if the epoch value is "unknown". It uses
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
+governed by the ReferenceTime type (optionally included in the "reference_time"
+field contained in a trace's "common_fields" ({{common-fields}})) and TimeFormat
+type (optionally included in the "time_format" field contained in the event
+itself, or a trace's "common_fields").
+
+There is no requirement that events in the same trace use the same time format.
+However, using a single time format for related events can make them easier to
+analyze.
+
+The reference time is defined as:
```
rmarx, line 681:
```suggestion
The reference time governs from which point in time the "time" field values are measured and is defined as:
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
-- set the reference_time to 1500 in "common_fields"
-- use: 0, 5, 22, 88
-~~~~~~~~
-{: #time-format-ex title="Three different approaches for logging timestamps"}
+The optional "wall_clock_time" field can be used to provide an approximate
+date/time value that logging commenced if the epoch value is "unknown". It uses
+the format defined in {{!RFC3339}}. Note that conversion of timestamps to
+calendar time based on wall clock times cannot be safely relied on.
+
+The time format is defined as:
```
rmarx, line 714:
```suggestion
The time format details how "time" values are encoded relative to the reference time and is defined as:
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
-separate post-processor to order qlog events in time if a tool do not provide this
-feature.
+order (though not necessarily absolute value, for the "relative_to_previous_event"
+format). Tools MAY sort all events on the timestamp before processing them,
+though are not required to (as this could impose a significant processing
+overhead). This can be a problem especially for multi-threaded and/or streaming
+loggers, who could consider using a separate post-processor to order qlog events
+in time if a tool do not provide this feature.
+
+Tools SHOULD NOT assume the ability to derive the absolute timestamp of an event
```
rmarx, line 739:
```suggestion
Tools SHOULD NOT assume the ability to derive the absolute calendar timestamp of an event
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -659,68 +666,186 @@ Example qlog event:
...
+~~~
+{: #rel-last-event-time-ex title="Relative-to-previous-event timestamps"}
+
+Example of a monotonic log using the relative_to_epoch format:
+
+~~~
+"common_fields": {
+    "time_format": "relative_to_epoch",
+    "reference_time": {
+          "clock_type": monotonic,
```
rmarx, line 820:
```suggestion
          "clock_type": "monotonic",
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -874,18 +999,24 @@ per-event instance:
...
     "events": [{
             "group_id": "127ecc830d98f9d54a42c4f0842aa87e181a",
-            "protocol_types": ["QUIC","HTTP/3"],
-            "time_format": "relative",
-            "reference_time": 1553986553572,
+            "protocol_type": ["QUIC","HTTP/3"],
+            "time_format": "relative_to_epoch",
+            "reference_time": {
+              "clock_type": "system",
+              "epoch": "2019-03-39T:22:55:53.572Z"
```
rmarx, line 1006:
```suggestion
              "epoch": "2019-03-29T:22:55:53.572Z"
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -874,18 +999,24 @@ per-event instance:
...
         },{
             "group_id": "127ecc830d98f9d54a42c4f0842aa87e181a",
-            "protocol_types": ["QUIC","HTTP/3"],
-            "time_format": "relative",
-            "reference_time": 1553986553572,
+            "protocol_type": ["QUIC","HTTP/3"],
+            "time_format": "relative_to_epoch",
+            "reference_time": {
+              "clock_type": "system",
+              "epoch": "2019-03-39T:22:55:53.572Z"
```
rmarx, line 1018:
```suggestion
              "epoch": "2019-03-29T:22:55:53.572Z"
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -900,9 +1031,12 @@ extracted to common_fields:
...
     "common_fields": {
         "group_id": "127ecc830d98f9d54a42c4f0842aa87e181a",
-        "protocol_types": ["QUIC","HTTP/3"],
-        "time_format": "relative",
-        "reference_time": 1553986553572
+        "protocol_type": ["QUIC","HTTP/3"],
+        "time_format": "relative_to_epoch",
+        "reference_time": {
+            "clock_type": "system",
+            "epoch": "2019-03-39T:22:55:53.572Z"
```
rmarx, line 1038:
```suggestion
            "epoch": "2019-03-29T:22:55:53.572Z"
```

rmarx (changes requested):
In general, this looks good :) 

I took a stab at the "millisecond resolution" part, but feel free to rework that a bit more if you feel it's needed.

Most comments here are nits/clarifications; in general this looks ready to! It's about time too ;)

draft-ietf-quic-qlog-main-schema.md
LPardue: I don't think we need this, its profiling the RFC even further.

draft-ietf-quic-qlog-main-schema.md
LPardue: I don't understand this bit. If I have a float64 field, and I decide to use it to log "1.5" to mean one and a half seconds, but the logger tool thinks I meant one and a half milliseconds, then there's an interop issue. So I think we need to pick one form of "units of x with fractional Y", or add a field to articulate the concrete units and precision.

PR #438, opened 2024-10-18
Add unknown_parameters field

### LINKED ISSUE #436: Logging unknown TransportParameters
Back in https://github.com/quicwg/qlog/issues/176 I said 

> The spec would be improved by having a well-defined "extension bucket" field that TPs can go in, an array of integers tuples would be fine, for example.

and we addressed that with a structured solution using group sockets in https://github.com/quicwg/qlog/pull/417.

However, I think one original problem still stands. As a receiver, if I receive an unknown TP, there's not a standard way to log the codepoint and value. H3 already has a solution to this via the SETTINGS frame but QUIC TPs are different and it isn't so straightforward. 

@hawkinsw has this problem right now. He is doing research on the Accurate ECN extension and very kindly proposed some [en/decode to quiche](https://github.com/cloudflare/quiche/pull/1860). And while a small patch, adding code for an I-D we don't implement (and might want to track all I-D revisions that change the code point) adds overhead. So it got me wondering could we use qlog instead and require zero changes, leading back here.

For example, by extending parameters_set as below, we could easily externalise the code that checks supported TPs to log tooling

```
QUICParametersSet = {
  // all the existing stuff
  
  unknown_parameters: [* UnknownTransportParameter]
  
}

UnknownTransportParameter= {
    id: uint64,
    value: uint64
}
```

rmarx:
Closes #436 

I decided not add something similar to `h3:parameters_set` since it makes more sense to just log it as H3Settings there, and that's already possible.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -664,6 +664,10 @@ remote peer). This is reflected in the `owner` field. As such, this field MUST b
 correct for all settings included a single event instance. If you need to log
 settings from two sides, you MUST emit two separate event instances.
 
+The `unknown_parameters` field can be used to log the raw values of any unknown
+transport parameters (e.g., GREASE, private extensions, peer-side
+experimentation).
+
 In the case of connection resumption and 0-RTT, some of the server's parameters
 are stored up-front at the client and used for the initial connection startup.
 They are later updated with the server's reply. In these cases, utilize the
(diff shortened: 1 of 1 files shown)


rmarx:
PTAL @hawkinsw @LPardue , I think this covers the load :)

LPardue (approved):

hawkinsw:
Sorry I was slow getting mine ready. I just added another PR that adds unknown parameters in a few additional places. If you would prefer to keep it to just the `parameter_set`, then that's great. I fixed another small typo in my PR. So, if you do end up accepting this PR rather than mine, I will open another one with just that typo fix. Sorry again for not getting this done sooner. I'm really disappointed in myself.

Issue #432, opened 2024-07-15
Extensions to events can not be indicated by schema


LPardue:
Here's a summary of what qlog already permits

* Define new event schema
  * Define new namespaces, if desired 
* Define new categories
  * In existing namespaces, if desired 
* Define new event types in new categories
  * Existing categories cannot be extended
* Extend any event that allows it with new fields

For example, the QUIC packet_sent event is extensible. Cloudflare has a custom field called send_at_time that we use (see https://github.com/quicwg/qlog/issues/297 for background). If I want to formally define that extension point with a schema, the current event_schemas mechanism doesn't quite work well. While I could define a new namespace, category ID and event type - that's a lot of overhead for adding just a single field to an existing event. This also makes it hard to compose extensions together.

Proposal: define a new type of schema that is scoped to cover only extensions to existing event types, which is indicated by a value in the URI e.g., where we have `urn:ietf:params:qlog:events:gen#loglevel` a standardized extension `foo` could be identified by `urn:ietf:params:qlog:event_ext:extension_foo`, a private extension could be identified by `https://example.com/032024/event_ext/extension_foo.html`

rmarx:
After my long comment above and thinking about it some more + a long discussion with Lucas, we decided it would make most sense to just remove categories from qlog altogether.

I originally used categories because I thought events would be re-usable across protocols, but in practice, this was a pipe dream anyway. They just made extensibility and naming (and even implementations) more difficult while adding little to no value. 

#439 removes categories and consequently makes it possible to be more flexible in how we extend qlog, including with new documents that just extend existing events in existing namespaces, and it thus also closes this issue as solved.

PR #439, opened 2024-10-18
Remove categories

rmarx:
As discussed during today's editors meeting with @LPardue, I feel it would be better and simpler overall to just remove the notion of event categories and only have event `namespaces` to group event types by. 

This simplifies handling extensions and schema URIs considerably, while also allowing us to more easily extend existing namespaces with new events (e.g., QUIC extensions can be properly tagged as `quic:my_new_event` instead of `my_quic_extension:my_new_event` which would otherwise have been needed).

Categories were originally needed because I had the vision of reusable event types across protocols (hence stuff in the `connectivity` or `security` categories would be re-usable across both TPC and QUIC for example). In practice, this has turned out to be a pipe dream and the QUIC document has also moved away from that in all but name (and the HTTP/3 document just had a single category anyway). 

(note: Friday 18 oct 4PM CEST: not fully ready yet; still need to remove the categories notion from QUIC and H3 docs, but the main doc text should be ready for review already)

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -77,27 +77,27 @@ containing concrete events for the core HTTP/3 protocol {{RFC9114}} and selected
 extensions ({{!EXTENDED-CONNECT=RFC9220}}, {{!H3_PRIORITIZATION=RFC9218}}, and
 {{!H3-DATAGRAM=RFC9297}}).
 
-The event schema namespace `http` is defined, containing the category `h3`; see
-{{schema-def}}. In this category multiple events derive from the qlog abstract
-Event class ({{Section 7 of QLOG-MAIN}}),  each extending the "data" field and
-defining their "name" field values and semantics
+The event namespace with identifier `http3` is defined; see {{schema-def}}. In
+this namespace multiple events derive from the qlog abstract Event class
+({{Section 7 of QLOG-MAIN}}), each extending the "data" field and defining
+their "name" field values and semantics.
 
 {{h3-events}} summarizes the name value of each event type that is defined in
-this specification. Some event data fields use complex datastructures. These are
+this specification. Some event data fields use complex data types. These are
 represented as enums or re-usable definitions, which are grouped together on the
 bottom of this document for clarity.
 
-| Name value                | Importance |  Definition |
-|:--------------------------|:-----------|:------------|
-| h3:parameters_set         | Base       | {{h3-parametersset}} |
-| h3:parameters_restored    | Base       | {{h3-parametersrestored}} |
-| h3:stream_type_set        | Base       | {{h3-streamtypeset}} |
-| h3:priority_updated       | Base       | {{h3-priorityupdated}} |
-| h3:frame_created          | Core       | {{h3-framecreated}} |
-| h3:frame_parsed           | Core       | {{h3-frameparsed}} |
-| h3:datagram_created       | Base       | {{h3-datagramcreated}} |
-| h3:datagram_parsed        | Base       | {{h3-datagramparsed}} |
-| h3:push_resolved          | Extra      | {{h3-pushresolved}} |
+| Name value                   | Importance |  Definition |
+|:-----------------------------|:-----------|:------------|
+| http3:parameters_set         | Base       | {{h3-parametersset}} |
+| http3:parameters_restored    | Base       | {{h3-parametersrestored}} |
+| http3:stream_type_set        | Base       | {{h3-streamtypeset}} |
+| http3:priority_updated       | Base       | {{h3-priorityupdated}} |
+| http3:frame_created          | Core       | {{h3-framecreated}} |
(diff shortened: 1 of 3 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -949,103 +930,114 @@ fields are "time" and "data", who are divergent by nature.
...
-two categories in the generic event schema defined in this document
-({{generic-event-schema}}), to three categories in an event schema named `rick`
-specified in a hypothetical RFC, and to three more categories from a private
-event schema named pickle. The standardized category URIs use a URN format, the
-private categories use a URI with domain name.
+Concrete event types (for example: packet_sent), as well as related data types
+(for example: QUICPacketHeader), are grouped in event namespaces (for example:
+quic) which in turn are defined in one or multiple event schemas (for example:
+urn:ietf:params:qlog:events:quic and
+urn:ietf:params:qlog:events:quic#extension1).
```
LPardue, line 937: nit: I find the inband examples break up the flow, so perhaps consider an example underneath

rmarx:
This PR is now 99% ready (I just need to add the full list of QUIC events to the IANA registration in the QUIC document), so also ready for full review.

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -71,14 +71,13 @@ containing concrete events for the core QUIC protocol (see
...
-The event schema namespace `quic` is defined, containing the categories:
-`connectivity` ({{conn-ev}}), `security` ({{sec-ev}}), `quic` {{quic-ev}}, and
-`recovery` {{rec-ev}}. Across these categories multiple events derive from the
-qlog abstract Event class ({{Section 7 of QLOG-MAIN}}), each extending the
-"data" field and defining their "name" field values and semantics. Some data
-fields use complex datastructures. These are represented as enums or re-usable
-definitions, which are grouped together on the bottom of this document for
-clarity.
+
+The event namespace with identifier `http3` is defined; see {{schema-def}}. In
```
LPardue, line 75:
```suggestion
The event namespace with identifier `quic` is defined; see {{schema-def}}. In
```

draft-ietf-quic-qlog-main-schema.md
```diff
@@ -949,103 +930,114 @@ fields are "time" and "data", who are divergent by nature.
...
+types they contain in the "qlog event schema URIs" registry.
 
-Private or non-standard event categories MAY be registered in the "qlog event
-category URIs" registry but MUST NOT use a URN of the form
-`urn:ietf:params:qlog:events:<namespace identifier>#<category identifier>`. URIs
+Event schema that define a new namespace SHOULD use a URN of the form
+`urn:ietf:params:qlog:events:<namespace identifier>`, where `<namespace
+identifier>` is globally unique. For example, this document defines two event
+schemas ({{generic-event-schema}}) for two namespaces: `loglevel` and `sim`.
+Other examples of event schema define the `quic` {{QLOG-QUIC}} and `http`
```
LPardue, line 1014:
```suggestion
Other examples of event schema define the `quic` {{QLOG-QUIC}} and `http3`
```

LPardue:
I tried an implementation of this, it reduced code and made it more logical to call. I support the change.

The one question I have, is #432 resolved by this? An event schema that only extends existing event types with new fields, will have an empty `Event Types` column in the IANA registration. Should we think about adding a column to make it explicit what the new fields are? That seems a bit tedious to manage, especially since we have many extension points through the documents. 

I think this PR finds the right balance of allowing extension schema identification, while avoiding having to import every detail to IANA.

rmarx:
@LPardue thanks for trying it out in actual implementation! Good to know this actually works as intended :) 

I do think this solves #432, as you'd just register the new schema without any concrete Event Types. The reason why the Event Types are now also in IANA is because we can now extend existing namespaces with new events, so we need to make sure new events don't collide with older events of the same name (put differently: now only the concatenation of `namespace:event_type` is globally unique across documents). Since extensions to existing events/data types can't collide with previous definitions, they don't need to be in IANA (imo). 

As such, I also think this approach indeed strike the right balance between ensuring interop and practicality.

I've resolved your comments above + added the event_types list in the IANA section. PTAL and let me know if this is ready to merge later today :) 

@marten-seemann if you get a moment today, it would be good to have 1 extra pair of eyes on this, since it's a pretty substantial departure from the previous setup.

draft-ietf-quic-qlog-h3-events.md
```diff
@@ -743,10 +746,16 @@ document as well.
 
 # IANA Considerations
 
-This document registers a new entry in the "qlog event category URIs" registry.
+This document registers a new entry in the "qlog event schema URIs" registry.
```
LPardue, line 749: missing the `Description` field

LPardue: oh it was below, resolved

LPardue (approved):
Ship it!

PR #440, opened 2024-10-18
Add support for logging unknown parameters

### LINKED ISSUE #436: Logging unknown TransportParameters

hawkinsw:
Add support for logging unknown parameters (including parameters set during connection recovery and parameters restored during 0-RTT).

Closes #436.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -670,6 +670,11 @@ They are later updated with the server's reply. In these cases, utilize the
 separate `parameters_restored` event to indicate the initial values, and this
 event to indicate the updated values, as normal.
 
+Implementations are not required to recognize every setting received in all situations.
+For example, QUIC implementations MUST discard transport parameters that they do not
+understand {{Section 7.4.2 of QUIC-TRANSPORT}}. `unknown_parameters` MAY be used to store
+the parameters from a peer that an implementation does not recognize, process or support.
+
 ~~~ cddl
 QUICParametersSet = {
     ? owner: Owner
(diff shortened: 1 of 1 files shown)


rmarx:
@hawkinsw thanks for this! As you saw, I had already made a stab at this at #438. Yours had some good things in it, so I merged them in my PR and will close this one. 

I also did drop a few things from this though, mostly due to how we do qlog in general + the logic behind restored parameters:
- in qlog, length fields are implicit if the value is logged. So for UnknownParameter, I just have the value as a hexstring, not the separate length field
- Unknown parameters are unlikely (or even impossible?) to restore either as part of 0-RTT or congestion control, seeing as we don't know what to do with them locally and restoration is only for local logic (restored parameters are not communicated to the other endpoint, just used to bootstrap the local one pending updates from the other). So I don't have these in the two events you proposed. 

Thanks again for your help on this!

PR #444, opened 2024-11-04
Indicate send blocking in data_moved events

### LINKED ISSUE #132: Add events to explicitly indicate send blocking
msquic has separate events that indicate if the sender was blocked/delayed/idle and due to which cause (e.g., pacing, flow control limited, cwnd limited, etc.).

This is quite useful for debugging. While some of this can be deduced from looking at other existing qlog events, adding some more explicit events for this could be interesting. 

@nibanks: maybe you could give some pointers/links to where these events are defined for msquic for reference? 

cc @lpardue

LPardue:
Fixes #132 (if merged).

It struck me that its the "data_moved" action that can be blocked, so rather than add
a separate event, just add an optional field to an existing event. This helps ensure we
can represent both stream and datagram data with all relevant fields.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1200,6 +1200,12 @@ processed first and afterwards the application layer reads from the streams with
 newly available data). This can help identify bottlenecks, flow control issues,
 or scheduling problems.
 
+The `blocked_reason` field supports optional logging of information related to
+stream writes getting blocked, either in full or in part. For example, if an
+application intended to write data to a stream but was blocked from sending
+anything due to congestion control, it can log the `congestion_control` reason
+and omit the length field.
+
 The `additional_info` field supports optional logging of information
 related to the stream state. For example, an application layer that moves data
 into transport and simultaneously ends the stream, can log `fin_set`. As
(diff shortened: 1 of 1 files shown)


rmarx:
Conceptually this works, but I don't like the suggestion we'd have to omit logging `length` to indicate the data was actually not moved... in practice, length could be a strong signal that helps understand why exactly something was blocked for that reason (e.g., trying to write a ridiculous amount of data due to a bug).

If we keep this approach, I'd rather the guidance for this is "if there's a `blocked_reason`, that means the data was not actually moved, but a move was attempted, which was blocked due to `blocked_reason`". That imo also works :)

LPardue:
Fair comment. For context, the quiche implementation has a stream write operation that takes an arbitrarily large buffer and can return a "partial write" result. For example, if the cwnd is 15 KB, and a 20 KB buffer is provided, well return an integer length result of 15 KB, indicating the app was blocked from writing the last 5 KB. 

I was thinking it would be nice to reflect that in the qlog as a stream_data_moved with length 15 KB and blocked_reason of congestion window, etc. 

I didn't think through everything in this proposal and maybe more guidance is useful, but it felt at the time of writing that combing the data in a single event would be more convenient for my use than if I had two separate events. 

Always happy to hear other opinions or review alternatives though.

nibanks:
My personal preference is for separate events. We already have it built this way in msquic.

Also, I agree with Kazuho. We don't have any notion of moving data between layers.

LPardue:
Ack. Would someone be willing to help on the alternative proposal? Even if ypu can just frame how the alternative works, we can design the event structure definition around that.

LPardue:
Conversation moved back to issue

PR #446, opened 2025-03-01
Add event_schemas to examples

mengelbart:
According to [section 3](https://datatracker.ietf.org/doc/html/draft-ietf-quic-qlog-main-schema-10#section-3), event_schemas is required.

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -347,6 +347,7 @@ JSON serialization example:
     "serialization_format": "application/qlog+json",
     "title": "Name of this particular qlog file (short)",
     "description": "Description for this group of traces (long)",
+    "event_schemas": ["urn:ietf:params:qlog:events:quic"],
     "traces": [...]
 }
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -492,6 +493,7 @@ JSON-SEQ serialization example:
     "serialization_format": "application/qlog+json-seq",
     "title": "Name of JSON Text Sequence qlog file (short)",
     "description": "Description for this trace file (long)",
+    "event_schemas": ["urn:ietf:params:qlog:events:quic"],
```
rmarx, line 496: Since the `protocol_types` field below is for both QUIC and H3, it would make sense for this to also include the HTTP/3 schema URI:

```suggestion
    "event_schemas": ["urn:ietf:params:qlog:events:quic", "urn:ietf:params:qlog:events:http3"],
```

rmarx (changes requested):

rmarx:
Thanks for this @mengelbart!

Though boy, did you open a can of worms in my head :D (see https://github.com/quicwg/qlog/issues/448).

For this specific PR though, I just had 1 thought that I've added as a suggestion above.

mengelbart:
Glad I could help :D 
I'm going to press "commit suggestion", just to make sure, but I assume this one can be closed because it's included in #449, right?

rmarx:
This was indeed done in #449, which was now merged with attribution, so I'm closing this PR :) thanks again!

PR #450, opened 2025-03-06
Be consistent in how we allow logging of raw values

### LINKED ISSUE #421: Make logging raw byte values consistent
As detailed here https://github.com/quicwg/qlog/pull/383#issuecomment-2194521047, we're not really consistent in how we allow logging of raw byte values of individual fields, especially if they're "unknown"/don't map to a known string or name or type or...

This is a reminder to @rmarx to make a PR that makes this consistent across all events that currently don't use the `field_bytes: uint64` (or similar) pattern.

### LINKED ISSUE #447: PING frame doesn't have a payload
In https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#section-8.13.2

```
 PingFrame = {
    frame_type: "ping"

    ; total frame length, including frame header
    ? length: uint32
    ? payload_length: uint32
}
```

In QUIC, the [PING frame](https://datatracker.ietf.org/doc/html/rfc9000#section-19.2) doesn't have a payload so the payload length can never be populated. In my implementation, I've always set this to `None` and the fields are always omitted.

Proposal: drop both the `length` and `payload_length` fields

rmarx:
- Use RawInfo instead of separate length/payload_length fields + make sure RawInfo is defined everywhere (fixes #447)
- Make sure that we have a separate `*_bytes` field to log raw values for unknown frames/errors/packet_types/... (fixes #421)

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1891,16 +1891,13 @@ such, each padding byte could be theoretically interpreted and logged as an
 individual PaddingFrame.
 
 However, as this leads to heavy logging overhead, implementations SHOULD instead
-emit just a single PaddingFrame and set the payload_length property to the amount
-of PADDING bytes/frames included in the packet.
+emit just a single PaddingFrame and set the raw.payload_length property to the
+amount of PADDING bytes/frames included in the packet.
 
 ~~~ cddl
 PaddingFrame = {
     frame_type: "padding"
-
-    ; total frame length, including frame header
-    ? length: uint32
-    payload_length: uint32
+    ? raw: RawInfo
 }
 ~~~
 {: #paddingframe-def title="PaddingFrame definition"}
@@ -1910,10 +1907,7 @@ PaddingFrame = {
 ~~~ cddl
 PingFrame = {
     frame_type: "ping"
-
-    ; total frame length, including frame header
-    ? length: uint32
-    ? payload_length: uint32
+    ? raw: RawInfo
 }
 ~~~
 {: #pingframe-def title="PingFrame definition"}
@@ -1942,10 +1936,7 @@ AckFrame = {
     ? ect1: uint64
     ? ect0: uint64
     ? ce: uint64
-
-    ; total frame length, including frame header
-    ? length: uint32
-    ? payload_length: uint32
+    ? raw: RawInfo
 }
 ~~~
 {: #ackframe-def title="AckFrame definition"}
@@ -1967,10 +1958,7 @@ ResetStreamFrame = {
 
     ; in bytes
     final_size: uint64
-
-    ; total frame length, including frame header
-    ? length: uint32
-    ? payload_length: uint32
+    ? raw: RawInfo
 }
 ~~~
 {: #resetstreamframe-def title="ResetStreamFrame definition"}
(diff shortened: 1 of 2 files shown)


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1958,35 +1965,46 @@ that packet with number 120 was ACKed). However, in that case, implementers SHOU
...
+error_code_bytes field; a numerical value without variable-length integer
+encoding.
+
 ~~~ cddl
 ResetStreamFrame = {
     frame_type: "reset_stream"
     stream_id: uint64
-    error_code: $ApplicationError /
-                uint64
+    error_code: $ApplicationError
```
LPardue, line 1978: So for my library, the QUIC layer is responsible for generating qlog frames and it is not application aware. That means I always just send the numerical code right now in `error_code`. Requiring me to send "unknown" will inflate my logs without any real benefit. Can we consider making `error_code` optional if going this route?

draft-ietf-quic-qlog-quic-events.md
```diff
@@ -1958,35 +1965,46 @@ that packet with number 120 was ACKed). However, in that case, implementers SHOU
...
+error_code_bytes field; a numerical value without variable-length integer
+encoding.
+
 ~~~ cddl
 StopSendingFrame = {
     frame_type: "stop_sending"
     stream_id: uint64
-    error_code: $ApplicationError /
-                uint64
+    error_code: $ApplicationError
```
LPardue, line 2002: Similar comment here as https://github.com/quicwg/qlog/pull/450/files#r1986392613

draft-ietf-quic-qlog-quic-events.md
rmarx: Gosh, I'm not the biggest fan of this... the main reason to do this type of change (imo) is so it's easier what to look at when parsing a qlog. With this, you can always assume the `error_code` is a string and just use that, and only if the value is `unknown` do you need to maybe look for another field. If you make `error_code` optional and people have to look for the optional presence of `error_code_bytes` as well, then there's not that much benefit to changing this from `$ApplicationError / uint64` to two different fields, imo

LPardue: What do you mean by "look at when parsing"? My parser attempts to parse all the defined fields statelessly. It's autogenerated code that currently can determine the qlog type based on whether its parsing a JSON string or number. Since JSON doesn't require order of fields, to be maintained, I'm not sure how a parser can do much statefully without already having parsed the object.

If you mean tools dealoling with the qlog object, then I can follow that argument. However, I just render the fields that are in the log, rather than do any checking or enforcement.

I can live with the current definition, it's straightforward and avoids weird scenarios where a well known error code is logged simply ad an integer rather than a string claiming "unknown", which is confusing IMO.

LPardue: In that case, I don't think it's too onerous on a more-capable tool to allow for error_code to be optional. It can just treat an omitted field as the default "unknown"

LPardue: another option would be for me to log an empty string (which I think is inherently allowed) - that doesn't require any change to this PR and lets me shave some unnecessary/mislabelled bytes of `"unknown"`

rmarx: If that's ok for you, then that works for me as well :)

Issue #429, opened 2024-07-09
How is "trigger" a common subfield of "data"?


LPardue:
In https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#section-8.2-5 we say:

Any generic key-value map type can be assigned to $ProtocolEventData (the only common "data" subfield defined in this document is the optional trigger field, see [Section 7.4](https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema.html#trigger-field)).

I don't quite understand how this is a common field, since there is no CDDL for it. Should we be defining `$ProtocolEventData` with the actual optional `?trigger` field? Or is is more like "common by convention" or some other magic. 

Or is it actually the case that "trigger" is a design pattern that concrete event definitions can adopt but that they must explicitly declare the trigger field in their event definition?

LPardue:
FWIW: Of this 40 events we define for QUIC and H3, only 10 of them currently define a trigger field.

rmarx:
It's impossible to add the trigger field to `$ProtocolEventData` since new events aren't really inheriting from `$ProtocolEventData` but rather just adding themselves to it as a list of all possible events that can be used (put differently: `$ProtocolEventData` is an extensible ENUM, not a base class).

We don't really have a base class for events at this point, and so trigger is left as a bit of a loosely defined thing (more like as you say "a design pattern that concrete event definitions can adopt but that they must explicitly declare the trigger field in their event definition").

Given that we only have the one `trigger` field that would be common and that it's optional and not super-widespread, I personally think this is fine?

LPardue:
Thanks for the explanation, that makes more sense. I think we can probably wordsmith the text on triggers a bit. I just reread section 8.2 and 8.2.1 and its seems a bit verbose. I think we can probably simplify this to something along the lines of "if its useful for an event to include a trigger, then it is RECOMMENDED that the event data contains an optional field named `trigger` that has a string value."

PR #451, opened 2025-03-12
Editorial work of triggers

LPardue:
As discussed on https://github.com/quicwg/qlog/issues/429, it seems like a better option to just RECOMMEND a "trigger" field if an event would benefit from an explicit trigger.

This change removes the mention of triggers from the parent section (8.2) and puts the normative sugestion in the trigger section (8.2.1)

Closes https://github.com/quicwg/qlog/issues/429

### DIFF
--- a/draft-ietf-quic-qlog-main-schema.md
+++ b/draft-ietf-quic-qlog-main-schema.md
@@ -1317,7 +1317,7 @@ $$quic-packetsent-extension //= (
 
 ### Triggers {#trigger-field}
 
-It can be is useful to understand the cause or trigger of an event. Sometimes,
+It can be useful to understand the cause or trigger of an event. Sometimes,
 events are caused by a variety of other events and additional information is
 needed to identify the exact details. Commonly, the context of the surrounding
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-main-schema.md
```diff
@@ -1322,36 +1317,44 @@ $$quic-packetsent-extension //= (
...
-This choice was made because many event types do not include a trigger
-value, and having the field at the Event-level would cause overhead in
-some serializations. Additional information on the trigger can be added
-in the form of additional member fields of the "data" field value, yet
-this is highly implementation-specific, as are the trigger field's
-string values.
-
-One purely illustrative example of some potential triggers for QUIC's
-"packet_dropped" event is shown in {{trigger-ex}}:
+It can be is useful to understand the cause or trigger of an event. Sometimes,
```
rmarx, line 1320:
```suggestion
It can be useful to understand the cause or trigger of an event. Sometimes,
```

rmarx (approved):
Minus the nit, LGTM!

Issue #79, opened 2020-04-12
Add support for connection migration


rmarx:
Now that this is a bit more mature, it should be added to draft-02.

Via @marten-seemann:
> So I would assume that would best be modeled using a `path_probed` event (which has ips, ports, CID), and then a `path_confirmed` and a `path_abandoned` event

Somewhat ties into #49 and #78 (and all other connectivity events really...).
Also look into how this is modeled in qlog for multipath by UCL (@mpiraux)

LPardue:
@rmarx was this fixed by https://github.com/quicwg/qlog/pull/336

rmarx:
Indeed, this can be closed :)

PR #453, opened 2025-03-12
Update QUICConnectionStarted event

### LINKED ISSUE #119: Make initiator clear in connection_started
Currently, `connection_started` utilizes `src_ip` and `dst_ip` (and equivalent for ports), where the intended meaning is that src means the initiator (typically client) and dst the server. 

However, this is a) unclear and b) it does not allow deducing what the "local" ip is for where this trace was captured without correlating client/server with the vantagepoint (and if that vantagepoint is network, it becomes more murky). 

Some options for solving:
1. rename fields to `local_ip` and `remote_ip`  
2. add `initiator` field (e.g., with values 'local' and 'peer')
3. completely revise how we log paths/IPs and potentially decouple from `connection_started` (also good in terms of privacy, migration, multipath, ...). 

Option 3 is probably best, should revisit when properly fixing migration in qlog (see also #79 and #57)

cc @marten-seemann

LPardue:
We've had issue #119 open for over 4 years. It proposed a few options
with the preference towards option 3, which related to #57 and #79.
Both of those issues have been fixed. In parallel, we've also got rid
of the notion of generic "transport" events and everything is framed
in terms of QUIC now.

More to the point, we added the PathEndpointInfo type that contains
all the same sorts of information that's needed, and it has text that
describes directionality aspects that #119 was concerned about.
Therefore, I lets just use that.

Also remove the `protocol` field now that the event is purely QUIC.

Closes #119

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -289,7 +289,7 @@ importance level; see {{Section 9.2 of QLOG-MAIN}}.
 QUICConnectionStarted = {
     local: PathEndpointInfo
     remote: PathEndpointInfo
-    
+
     * $$quic-connectionstarted-extension
 }
 ~~~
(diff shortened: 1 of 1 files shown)


rmarx:
I love it when things come together in the end :) Let's merge this!

rmarx (approved):

PR #454, opened 2025-03-12
Add aborted connection_close trigger

### LINKED ISSUE #147: How to log handshake cancelations?
I didn't think of this back when we defined the triggers for `connection_closed`, but this is happening **a lot** in production: When dialing a new peer, libp2p races connections to that peer on all advertised addresses and transports, and uses whichever connection returns first. All the other connection attempts are canceled.

It would be nice if there was a trigger "the user aborted this connection attempt before handshake completion".

LPardue:
We talked about this on #147 and there seemed to be agreement to jsut do it.

Closes #147


rmarx (approved):
LGTM!

PR #455, opened 2025-03-12
add trigger and reason to H3PriorityUpdated

### LINKED ISSUE #324: Trigger for H3PriorityUpdated
I am wondering if the new H3PriorityUpdated event (see #312) needs some specified `trigger` field values.

Conceptually, you can get the necessary context from other events surrounding this event (e.g., if you get a H3FrameParsed with a HEADERS containing a `priority` field preceding this, that's a clear indicator that's the trigger). However, there are cases where this is less clear (e.g., default initialization to a value different from the spec, local override due to configuration or other logic, merging of client and server-sent priorities for the same stream, ...)

We don't NEED to define any triggers here (`trigger` is always defined on event data, just not necessarily specced in qlog), but I'm wondering if some default values would be useful here @LPardue?

e.g., 
```
? trigger: "client_directive" / "server_directive" / "local_override" / "local_default" / "merge" / "merge_override"
```

LPardue:
implements my latest proposal on #324

Closes #324

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -338,11 +338,11 @@ HTTP3PriorityUpdated = {
     ? old: HTTP3Priority
     new: HTTP3Priority
 
-    ? trigger = "client_signal_received" /
-                "local" / 
+    ? trigger: "client_signal_received" /
+                "local" /
                 "other"
 
-    ? reason = "client_signal_only" /
+    ? reason: "client_signal_only" /
                "client_server_merged" /
                "local_policy" /
                "other"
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-h3-events.md
```diff
@@ -316,12 +316,16 @@ $HTTP3StreamType /=   "request" /
...
-QLOG-MAIN}}.
+The `priority_updated` event is emitted when the priority of a request stream or
+push stream is initialized or updated through mechanisms defined in
+{{!RFC9218}}. It has Base importance level; see {{Section 9.2 of QLOG-MAIN}}.
+
+There can be several reasons why a `priority_updated` occurs, and why a
+particular value was chosen. For example, the priority can be updated through
+signals received from client and/or server (e.g., in HTTP/3 HEADERS or
+PRIORITY_UPDATE frames) or it can be changed or overridden due to local
+policies. The The `trigger` and `reason` fields can be used to optionally
```
rmarx, line 327:
```suggestion
policies. The `trigger` and `reason` fields can be used to optionally
```

rmarx (approved):
There's definitely some overlap between the two here which makes it feel less "clean". 

However, being a Prioritization Enthusiast myself, I can definitely see the value of having a split into two fields, allowing for fine-grained info (especially since the fields are optional). 

Minus the superfluous `The`, LGTM!

PR #457, opened 2025-03-12
Remove excessive citation of importance level section

LPardue:
I don't think the importance level is _that important_ that we need
a deep link into the qlog main schema every time. Its a visual
distraction. So instead, highlight the convention in the convention
section and remove all these redundant links.


rmarx:
Yes, this was annoying me as well!

One questions is if you intentionally not removed this from the events in the main schema as well (e.g., https://quicwg.org/qlog/draft-ietf-quic-qlog-main-schema#section-9.1)? 

I can see an argument for keeping them intact in that doc but not in the others, but if it wasn't intentional, maybe remove them there as well. Otherwise LGTM!

rmarx (approved):

LPardue:
Leaving them in main schema was an accident, I'll fix that up

PR #458, opened 2025-03-15
make the IPv4 and IPv6 2-tuple in the PreferredAddress optional

marten-seemann:
According to section 18.2 of RFC 9000, the server MAY only send one (valid) address:
> Servers MAY choose to only send a preferred address of one address family by sending an all-zero address and port (0.0.0.0:0 or [::]:0) for the other family. IP addresses are encoded in network byte order.

In that case, it should be valid to just not log the invalid address.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -712,10 +712,10 @@ QUICParametersSet = {
 }
 
 PreferredAddress = {
-    ip_v4: IPAddress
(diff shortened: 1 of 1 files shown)


LPardue:
Making it optional seems fine, but can we leave the field order the same for consitency with other events that carry IPs

marten-seemann:
> Making it optional seems fine, but can we leave the field order the same for consitency with other events that carry IPs

Fixed for the other ones in #460.

PR #462, opened 2025-04-20
Added missing events to QuicEventData

DannyG-1952723:
I noticed that 3 events were missing from `QuicEventData` (`QUICMigrationStateUpdated`, `QUICMarkedForRetransmit`, and `QUICEcnStateUpdated`), so I added these. The security events (`QUICKeyUpdated` and `QUICKeyDiscarded`) were placed above the transport events, while the text has it the other way around, so I moved these so `QuicEventData` has the same order as the text.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -209,8 +209,6 @@ QuicEventData = QUICServerListening /
                 QUICConnectionStateUpdated /
                 QUICPathAssigned /
                 QUICMTUUpdated /
-                QUICKeyUpdated /
-                QUICKeyDiscarded /
                 QUICVersionInformation /
                 QUICALPNInformation /
                 QUICParametersSet /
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -227,11 +225,16 @@ QuicEventData = QUICServerListening /
...
+                QUICKeyUpdated /
+                QUICKeyDiscarded /
                 QUICRecoveryParametersSet /
                 QUICRecoveryMetricsUpdated /
                 QUICCongestionStateUpdated /
                 QUICLossTimerUpdated /
-                QUICPacketLost
+                QUICPacketLost /
+                QUICMarkedForRetransmit /
+                QUICEcnStateUpdated
```
rmarx, line 237:
```suggestion
                QUICECNStateUpdated
```

For better or worse, we're capitalising this in the other part of the doc, so let's stay consistent :)

rmarx (approved):
Thanks @DannyG-1952723 for catching this (not sure why we didn't... :))

Just a small nit for consistency and then I think this can be merged.

draft-ietf-quic-qlog-quic-events.md
DannyG-1952723: Oh yeah, my bad :)

PR #468, opened 2025-07-08
Tweak stream_state_updated

### LINKED ISSUE #374: Improve quic:stream_state_updated

quic:stream_state_updated:

  I think this needs to be clarified a little that an event logged here with both
  a `stream_id` field and a `stream_side` field is expressing a state for that
  component of the stream and from the perspective of the specified vantage point
  (also, what about the network vantage point?); whereas an event logged here
  with a `stream_id` field but no `stream_side` field is expressing a state for
  both stream components as a whole.

  If my interpretation here is correct, I think it would also help to note that
  the valid values of StreamState depend on the value *and* presence of
  stream_side.

LPardue:
This implements my suggestions from #374, adding a new optional
initiator field to supplement the stream_type field, along with
making the stream_side field mandatory.

Having just implemented a prototype pcap to qlog tool, that tool
needs to try and build an understanding of who is a QUIC server
and client, and in what direction messages are flowing. So even
an observer vantage point can determine and log `stream_side`.

Closes #374

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1059,6 +1059,11 @@ inferred from several types of frames going over the wire, but it's much easier
 to have explicit signals for these state changes. The event has Base importance
 level.
 
+While stream IDs encode the type of stream and its initiator, the optional
+`stream_type` and `initiator` fields can be used to provide a more-accessible
+form of the information.
+
+
 ~~~ cddl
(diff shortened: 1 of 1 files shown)


marten-seemann (changes requested):
Isn't this duplicate now? For example, if we're a client, and we're dealing with a client-initiated uni-directional stream, we know for a fact that it's a send stream.

My preference would be for removing `stream_side` altogether.

PR #478, opened 2025-10-08
split up `PacketHeader.flags`

### LINKED ISSUE #471: `PacketHeader.flags` should be replaced with individual values
`PacketHeader.flags` requires consumers to implement a packet header parser. It would be nicer if we actually spelled out the values, i.e.
```
? quic_bit: uint8
? spin_bit: uint8
? key_phase: int
? packet_number_length: uint8
```

I imagine most implementations won't be interested in logging most of these values. For example, I'm not going to even implement the `spin_bit` flag in quic-go, since quic-go doesn't support the spin bit. Similarly, there's no value in logging `quic_bit`, unless the greasing extension is implemented (and even then, it's not clear if there's a lot of value in logging it).

Open question: Should we log the key phase bit, or the key phase (the key phase is derived from the bit). I'd argue that we should log the key phase, similar to how we log the decoded and not the wire packet number.

marten-seemann:
Fixes #471.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1794,19 +1794,12 @@ $PacketNumberSpace /= "initial" /
 
 ## PacketHeader
 
-If the packet_type numerical value does not map to a known packet_type string,
-the packet_type value of "unknown" can be used and the raw value captured in the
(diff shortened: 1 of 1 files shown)


rmarx (changes requested):
Something to discuss :)

rmarx (changes requested):

PR #482, opened 2025-10-08
add support for draft-ietf-quic-reliable-stream-reset

marten-seemann:
I just wanted to see what it would take to add support for draft-ietf-quic-reliable-stream-reset. Turns out it's really easy, we just need one transport parameter and one frame definition.


rmarx:
IIRC, we decided a long while back we would not keep incorporating newer RFCs into the main qlog docs (only the fixed set we already have in there), so this would sadly have to be a an extension doc...

LPardue:
Please make an issue for this question, then we can raise that with the WG

LPardue:
Raised during IETF 124, support in the room for merging it, so lets do so

PR #493, opened 2025-10-14
Clarify logging of negotiated parameters and how to combine parameters_set values

### LINKED ISSUE #398: Difference between negotiated vs actually accepted/enabled TPs

rmarx:
Fixes #398 

First stab at a simpler solution for that issue (supersedes previous attempt #409).

I tried to keep as much nuance as possible while adding as little prose as possible + not adding extra design/format elements.

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -657,8 +657,8 @@ regularly be several instances of this event with different fields set.
 
 Note that some settings have two variations (one set locally, one requested by
 the remote peer). This is reflected in the `owner` field. As such, this field
-MUST be correct for all settings included a single event instance. If you need
-to log settings from two sides, you MUST emit two separate event instances. If
+MUST be correct for all settings included a single event instance. If the settings
+from two sides are required, they MUST be logged as two separate event instances. If
 the local peer decides to change its behavior based on remote peer's settings,
 a new event type can be used to reflect the outcome.
 


draft-ietf-quic-qlog-quic-events.md
```diff
@@ -655,10 +655,18 @@ Most of these settings are typically set once and never change. However, they
...
 regularly be several instances of this event with different fields set.
 
-Note that some settings have two variations (one set locally, one requested by the
-remote peer). This is reflected in the `owner` field. As such, this field MUST be
-correct for all settings included a single event instance. If you need to log
-settings from two sides, you MUST emit two separate event instances.
+Note that some settings have two variations (one set locally, one requested by
+the remote peer). This is reflected in the `owner` field. As such, this field
+MUST be correct for all settings included a single event instance. If you need
+to log settings from two sides, you MUST emit two separate event instances. If
```
LPardue: The use of "you" in this paragraph sticks out. There's nothing else like it in our qlog specs. I know it was in the old text but while we are here lets change it. Something like below perhaps

```suggestion
MUST be correct for all settings included a single event instance. If the settings
from two sides are required, they MUST be logged as two separate event instances. If
```

LPardue (approved):

PR #504, opened 2026-03-02
Formally define HTTP3ExtensionFrames, refactor order of frame defs, and add ORIGIN frame

### LINKED ISSUE #503: Support HTTP/3 ORIGIN extension
Following up on some late additions of extension support, one really simple thing we're missing is the ORIGIN frame defined in https://datatracker.ietf.org/doc/html/rfc9412

LPardue:
…nd add ORIGIN frame

While adding ORIGIN frame I noticed we'd omitted mentioning PRIORITY_UPDATE in the schema extension point, so this fixes that too and hence the reordering makes it a bit clearer what is base vs. extension

fixes https://github.com/quicwg/qlog/issues/503

### DIFF
--- a/draft-ietf-quic-qlog-h3-events.md
+++ b/draft-ietf-quic-qlog-h3-events.md
@@ -711,7 +711,7 @@ The ORIGIN frame is defined in {{ORIGIN}}.
 ~~~ cddl
 HTTP3OriginEntry = {
   origin_len: uint16
-  ? ASCII-Origin: text
+  ? ascii_origin: text
 }
 
 HTTP3OriginFrame = {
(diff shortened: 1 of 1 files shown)


draft-ietf-quic-qlog-h3-events.md
```diff
@@ -688,33 +719,23 @@ HTTP3Priority = text
...
+The ORIGIN frame is defined in {{ORIGIN}}.
 
 ~~~ cddl
-HTTP3ReservedFrame = {
-    frame_type: "reserved"
-    frame_type_bytes: uint64
-    ? raw: RawInfo
+HTTP3OriginEntry = {
+  origin_len: uint16
+  ? ASCII-Origin: text
```
rmarx: it's not a hard rule I guess, but up until now we've had all lowercase for field names, as well as undescores instead of dashes... so should be:


```suggestion
  ? ascii_origin: text
```

rmarx (changes requested):

draft-ietf-quic-qlog-h3-events.md
LPardue: Ah, that was a copy/paster from the RFC, thanks.

PR #508, opened 2026-06-11
Add new optional action field to frames_processed

### LINKED ISSUE #507: frames_processed event for receiving *and* sending?
I'm prototyping qmux and qlog almost "just works" without any modification. However, one issue I hit was that I was only logging information about QUIC frames in packet_received and packet_sent events. HTTP/3 frames didn't hit the issue because we have frame_created and frame_parsed events there. The asymmetry makes some sense, due to how H3 is running over a stream.

In qmux, QUIC framing is also running over a stream and there are not packets. One hack for logging frames is to make fake packet events. Another would be to define a qmux event, perhaps a qmux_record_received if we go with https://github.com/quicwg/qmux/pull/26 (or overload the packet_received event for the record). Another potential event is frames_processed https://quicwg.org/qlog/draft-ietf-quic-qlog-quic-events.html#name-frames_processed, taking some choice words from the spec

> Implementations have the opportunity to (selectively) log this type of signal without having to log packet-level details (e.g., in packet_received). Since for almost all cases, the effects of applying a frame to the internal state of an implementation can be inferred from that frame's contents, these events are aggregated into this single frames_processed event.

> In some implementations, it can be difficult to log frames directly, even when using packet_sent and packet_received events.

Well, I don't have packet details so this seems ideal! However, if I use the frames_processed event I can't disambiguate between frame creation and frame parsing. Maybe we could add an optional field to explicitly indicate which one of these it is?

LPardue:
A logger might not have any access to QUIC packets,
for example a QMux implementation that sends records rather than packets.

This commit adds a new optional `action` field that allows for explicity
stating if the frames are create or parsed.

Furthermore, although the existing `packet_numbers` field can help in situations
where the logger _can_ log packets but not include frame information,
since packet numbers are not globally unique, using them leaves an ambiguity (
determing who sent or received a packet might be racy). The action field helps
there too.

Fixes #507

### DIFF
--- a/draft-ietf-quic-qlog-quic-events.md
+++ b/draft-ietf-quic-qlog-quic-events.md
@@ -1166,26 +1166,29 @@ important, the `packet_received` event can be used instead.
 
 In some implementations, it can be difficult to log frames directly, even when
 using `packet_sent` and `packet_received` events. For these cases, the
-`frames_processed` event also contains the `packet_numbers` field, which can be
-used to more explicitly link this event to the `packet_sent`/`received events`.
-The field is an array, which supports using a single `frames_processed` event
-for multiple frames received over multiple packets. To map between frames and
-packets, the position and order of entries in the `frames` and `packet_numbers`
-is used. If the optional `packet_numbers` field is used, each frame MUST have a
-corresponding packet number at the same index.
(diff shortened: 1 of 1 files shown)


LPardue:
closing as fixed by an laternative approach

